PR #2, opened 2016-05-28
add groupBy and groupByTransitively

lukewestby:
I've added what I believe are proper implementations of what I would expect `groupBy` and `groupByTransitively` to look like. Seems like it could be particularly useful for lists of records with an accessor function, although any computed value will work. Thanks!

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -22,7 +22,7 @@ module List.Extra exposing ( last
   , foldl1, foldr1
   , scanl1, scanr, scanr1, unfoldr
   , splitAt, takeWhileEnd, dropWhileEnd, span, break, stripPrefix
-  , group, groupWhile, groupWhileTransitively, inits, tails, select, selectSplit
+  , group, groupBy, groupByTransitively, groupWhile, groupWhileTransitively, inits, tails, select, selectSplit
   , isPrefixOf, isSuffixOf, isInfixOf, isSubsequenceOf, isPermutationOf
   , notMember, find
   , elemIndex, elemIndices, findIndex, findIndices
@@ -50,7 +50,7 @@ module List.Extra exposing ( last
 @docs scanl1, scanr, scanr1, unfoldr, iterate
 
 # Sublists
-@docs splitAt, takeWhileEnd, dropWhileEnd, span, break, stripPrefix, group, groupWhile, groupWhileTransitively, inits, tails, select, selectSplit
+@docs splitAt, takeWhileEnd, dropWhileEnd, span, break, stripPrefix, group, groupBy, groupByTransitively, groupWhile, groupWhileTransitively, inits, tails, select, selectSplit
 
 # Predicates
 @docs isPrefixOf, isSuffixOf, isInfixOf, isSubsequenceOf, isPermutationOf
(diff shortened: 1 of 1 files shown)


mgold:
Is it necessary that the computed property be comparable, when all you need is equatable? Bools and union types, for example, can be equated but not compared. Or will the list be sorted by the key?

What's the behavior for non-consecutive matches? e.g. `groupBy .x [ { x = 1 }, { x = 2 }, { x = 1 } ]`? 

Alternatively, what if these returned a dictionary, which would justify the comparable?

I'm open to "I'll fix this simple issue and stick the API changes in another issue and we'll come back to them" or other triaging.

jvoigtlaender:
Performance of the implementations as currently given in the pull request will be very bad in certain cases. For the same reasons as https://github.com/elm-lang/core/issues/485. I don't think we should put such vulnerability into the library. Better use the "trick" from the comment over there (that also Haskell uses).

jvoigtlaender:
@mgold, a version of groupBy with Dict already exists in https://github.com/elm-community/dict-extra.

jvoigtlaender:
Also, I don't see why two functions are added here. Given the properties of `==` in Elm, the two functions as implemented will _always_ return the same results for the same inputs. That is, the `groupBy` and `groupByTransitively` here are semantically indistinguishable.

jvoigtlaender:
In summary, I'm in favour of only adding `groupBy`, giving it this type (no `comparable` in there):

``` elm
groupBy : (a -> b) -> List a -> List (List a)
```

and implementing it in a way that does not apply the function more than once on any given list element.

Given the discussion that took place over at https://github.com/elm-community/elm-list-extra/issues/38, @eeue56 may want to weigh in here.

lukewestby:
All excellent feedback, thanks folks! I can get rid of `groupByTransitively` and implement a version of `groupBy` that doesn't recompute every time. If the computation is to not result in a `comparable` I think the docs should at least warn the user that certain types (I'm thinking of `Date` specifically) will behave in unexpected ways.

jvoigtlaender:
About the warning concerning strange behaviors of `==` on certain types: Other functions in this library, like `group`, suffer from the same strange behavior, but don't come with a warning in the docs. I guess the same could be said about several functions in `core`.

jvoigtlaender:
Regarding `core`: For example, `List.member` will behave strangely at type `Date`, but doesn't warn the user in its documentation. I don't see why the case of `groupBy` here is different.

lukewestby:
I think the behavior should be documented for users _somewhere_, whether it's core or otherwise, but I can still agree to keeping consistency in this package. No warning it is. Thanks!

jvoigtlaender:
@lukewestby, do you want to follow up on this?

lukewestby:
@jvoigtlaender I'll make the requested changes tomorrow and ping you when ready

lukewestby:
so sorry this fell through, i'm going to get to it as soon as i can

4ZM:
What is the status of this? Ended up re-implementing it today since I needed it. Would be nice to get from the lib.

lukewestby:
I'm not gonna get to this, if someone else wants to pick it up go for it

PR #5, opened 2016-06-11
Add the uniqueBy function.

garetht:
This pull request adds `dropDuplicatesBy`, which allows a user-specified function to be computed to determine the duplicate value.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -7,7 +7,7 @@ module List.Extra exposing ( last
   , andMap, andThen
   , takeWhile
   , dropWhile
-  , unique
+  , unique, uniqueBy
   , replaceIf
   , setAt
   , remove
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Seems wrong to me that it has this type:

``` elm
dropDuplicatesBy : (comparable -> comparable) -> List comparable -> List comparable
```

Rather, shouldn't it have the following type?

``` elm
dropDuplicatesBy : (a -> comparable) -> List a -> List a
```

jvoigtlaender:
Also, calling `f first` twice is unnecessarily inefficient. Share these calls via a `let`-binding?

garetht:
Completely agree. Fixed and thanks for the feedback!

jvoigtlaender:
This may have to wait until the resolving of #7, because of the name change concerning `dropDuplicates` discussed there.

abadi199:
@garetht can you rebase this PR and rename the function to follow with the new name change from #7?

garetht:
@abadi199 Done.

PR #9, opened 2016-06-24
add the splitByListOfLengths function

patrickdet:
There might be another, non-obvious to me, way to do this but I had to build this function for a library I am writing and thought I would share it with you all. Maybe it's useful for someone else as well.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -22,7 +22,7 @@ module List.Extra exposing ( last
   , foldl1, foldr1
   , scanl1, scanr, scanr1, unfoldr
   , splitAt, takeWhileRight, dropWhileRight, span, break, stripPrefix
-  , group, groupWhile, groupsOfVariableLength, groupWhileTransitively, inits, tails, select, selectSplit
+  , group, groupWhile, groupsOfVarying, groupWhileTransitively, inits, tails, select, selectSplit
   , isPrefixOf, isSuffixOf, isInfixOf, isSubsequenceOf, isPermutationOf
   , notMember, find
   , elemIndex, elemIndices, findIndex, findIndices
@@ -65,7 +65,7 @@ module List.Extra exposing ( last
 @docs lift2, lift3, lift4
 
 # Split to groups of given size
-@docs groupsOf, groupsOfWithStep, groupsOfVariableLength, greedyGroupsOf, greedyGroupsOfWithStep
+@docs groupsOf, groupsOfWithStep, groupsOfVarying, greedyGroupsOf, greedyGroupsOfWithStep
 -}
 
 import List exposing (..)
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
Use splitAt inside?

Also: Need to export your function from the module, and to add it to one of the docs tag lines.

shmookey:
Handy function, but pretty long name! Instead of naming it after the operations it performs, you might consider naming it after what you get back. How about `chunks`?

The documentation also seems a bit wordy. Maybe something like:

_`chunks ns` takes `n` elements from a list for each `n` in `ns`, splitting the list into variably sized segments_

patrickdet:
Thanks for the feedback! I have changed everything accordingly

shmookey:
Nice work. One small nitpick - using backticks in comments to highlight references to code will make the rendered documentation clearer.

At the moment, we just have a plain text string:
`{-| Chunks ns takes n elements from a list for each n in ns`

Whereas something like this:
`{-| `chunks ns` takes `n` elements from a list for each `n` in `ns``

...will render more like this:
`chunks ns` takes `n` elements from a list for each `n` in `ns`

Much clearer, I think. :smile:

shmookey:
Not directly relevant to this PR (unless you want to implement it), but while I think of it: I suggest `chunks` instead of `chunk` in anticipation of one day someone asking for `chunk : Int -> List a -> List (List a)`. 

Edit: Ack. The hypothetical `chunk` already exists, it's called `groupsOf`, and there's a whole family of related functions. Perhaps your function should follow that naming convention rather than introducing a new term "chunk".

patrickdet:
@shmookey that seems like a good plan. i will rename `chunks` to `groupsOfVariableLength`

patrickdet:
i know `groupsOfVariableLength` is a long name again but it follows the `groupsOf` naming scheme more clearly which is nice

jvoigtlaender:
How about the shorter name "groupsOfLengths"?

mgold:
I'd expect

``` elm
groupsOfLength : Int -> List a -> List (List a)
groupsOfLengths : List Int -> List a -> List (List a)
```

Although the first case is probably quite common, it's a pity that it has to be a list of list instead of a list of tuples. (Making the latter possible requires a lot of additional complexity in the type checker.)

jvoigtlaender:
How about, instead:

``` elm
groupsOf : Int -> List a -> List (List a)
groupsOfVarying : List Int -> List a -> List (List a)
```

patrickdet:
I like `groupsOfVarying` so i renamed the function.

On a related note: I have some test cases for this function but `List.Extra` is not using `ElmTest`. Is there a reason for that? Is it just that no one has set it up yet or was there a deliberate choice against it?

mgold:
Long term, I think it's because no one's bothered to set it up. Unit tests are always a good idea. Short term, though, elm-test is being overhauled and you might find it prudent to wait a few weeks.

patrickdet:
so what's everyone's opinion? is this good to merge?

jvoigtlaender:
Can you avoid the shadowing of the rest variable? Also, the head variable is unused. Why not simply match against _ :: _ there?

mgold:
What about:

``` elm
groupsOfVarying' : List Int -> List a -> List (List a) -> List (List a)
groupsOfVarying' listOflengths list accu =
    case ( listOflengths, list ) of
        ( length::tailLengths, _::_ ) ->
            let
                ( head, tail ) =
                    splitAt length list
            in
                groupsOfVarying' tailLengths tail (head::accu)

        _ ->
            List.reverse accu

```

PR #10, opened 2016-08-10
Add indexed fold functions

elliotdickison:
I was missing the index parameter available in JavaScript's reduce and reduceRight. These changes add indexedFoldl and indexedFoldr to handle use cases where that parameter is helpful/necessary. These functions are identical to List.foldl and List.foldr except that the index of the current element being processed is passed to the reducer function along with the element itself.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -19,7 +19,7 @@ module List.Extra exposing ( last
   , filterNot
   , iterate
   , intercalate, transpose, subsequences, permutations, interweave
-  , foldl1, foldr1
+  , foldl1, foldr1, indexedFoldl, indexedFoldr
   , scanl1, scanr, scanr1, unfoldr
   , splitAt, takeWhileRight, dropWhileRight, span, break, stripPrefix
   , group, groupWhile, groupWhileTransitively, inits, tails, select, selectSplit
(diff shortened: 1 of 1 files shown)


jvoigtlaender:
You forgot to export the functions from the module.

I would swap the order of the first and second arguments to the step function.

For both functions, simpler implementations seem possible by just combining "indexedMap (,)" and foldl/foldr.

elliotdickison:
Woah thanks for the lightning fast response. Added the exports and switched the args (good call).

I did consider indexedMap + the fold functions, but went with the current implementation to avoid taking an extra pass over the list. The particular use-case that I first wrote these for involved fairly large lists, and I figured that the (albeit very slight) performance gain was worth the slightly more complex implementation under the hood.

jvoigtlaender:
For indexedFoldr you do have two traversals in the currrent version as well, due to the List.length call. But I do get that it may still be more efficient than going via indexedMap.

elliotdickison:
Interesting, I didn't know that (makes sense). I can't think of a way around it w/o dropping down to JS though (and even then it would be 2 traversals at best instead of the current 3, looking at the implementation of List.foldr). Thoughts?

jvoigtlaender:
Let's hear what the maintainer of this package thinks. @abadi199

abadi199:
I think the current implementation is fine. If anyone can think of a better implementation, just make a new pull request. I'll merge this and publish the package.

abadi199:
Published! http://package.elm-lang.org/packages/elm-community/list-extra/3.1.0/
Thanks @elliotdickison for your contribution.

elliotdickison:
Thanks @abadi199 !

PR #16, opened 2016-09-29
Add swapAt function.

roSievers:
This pull request adds the swapAt function to list-extra. It can swap two values in a list by index. Returns the updated list if both indices are in range, or Nothing if both are out of range. If the same index is supplied twice the original list is returned.

This could be implemented using getAt, setAt but that would deconstruct and reconstruct the list multiple times. It also requires more Maybe wrangling.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -10,6 +10,7 @@ module List.Extra exposing ( last
   , unique, uniqueBy
   , replaceIf
   , setAt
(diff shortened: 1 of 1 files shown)


abadi199:
Merged here: https://github.com/elm-community/list-extra/commit/ef6b42353e9be72c14ab5bc26d248ddc53e1258e

PR #19, opened 2016-11-10
Fix documentation for break

Menschenkindlein:
```
`span p xs` is equivalent to `(dropWhile p xs, takeWhile p xs)`
```
should be changed into
```
`break p xs` is equivalent to `(takeWhile (not p) xs, dropWhile (not p) xs)`
```


abadi199:
Thanks for the contribution 😃

PR #27, opened 2016-12-19
Rewrite takeWhile so it can be tail-call-optimised

TSFoster:
Hi,

takeWhile was crashing on long lists, rewriting it this way allows the elm compiler to use a loop in JS.


prikhi:
Thanks

PR #31, opened 2017-01-20
Add initialize function.

miguel-vila:
I was surprised to find out that there is [`initialize`](http://package.elm-lang.org/packages/elm-lang/core/latest/Array#initialize) for `Array`s but not for `List`s, as far as I know. I implemented it starting the computation from the end so it's tail recursive. Let me know your thoughts.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -27,6 +27,7 @@ module List.Extra
         , removeAt
         , filterNot
         , iterate
+        , initialize
         , intercalate
         , transpose
         , subsequences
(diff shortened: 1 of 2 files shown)


zwilias:
Alright, figured I'd also have a look at this.

Here's an alternative implementation that seems much easier to understand, uses only simply primitives and has very similar performance characteristics:

```elm
initialize : Int -> (Int -> a) -> List a
initialize size f =
   List.range 0 (size - 1) |> List.map f
```

https://ellie-app.com/3nSzNC5pmNqa1/0 for benchmarks.

Chadtech:
Yeah that looks a lot better. 

Maybe I was too quick to merge this PR. Evaluating the code I merged was difficult.

Chadtech:
Oh, @zwilias, I think there is a mistake in your ellie-app. You use `initialize` in both PR and Native. Thats why it has similar performance. I fixed your error and found that the PR code performs at least twice as well.

zwilias:
Oh, 😅

PR #33, opened 2017-01-23
`singleton` is now in core

toastal:
The [latest version of `core`'s `List` already contains `singleton`](http://package.elm-lang.org/packages/elm-lang/core/5.1.0/List#singleton). This causes a naming conflict when trying to do `import List.Extra as List`.

PR #34, opened 2017-01-23
Remove List.Extra.singleton because List.singleton now exists in core

jvoigtlaender:
http://package.elm-lang.org/packages/elm-lang/core/5.1.0/List#singleton


mgold:
Duplicate of #33.

Issue #41, opened 2017-04-25
Proposal for splitWhen function


EverybodyKurts:
It would combine the following function signatures:

```
findIndex : (a -> Bool) -> List a -> Maybe Int
splitAt : Int -> List a -> (List a, List a)
```

Here's a first stab at it:

```
splitWhen : (a -> Bool) -> List a -> Maybe ( List a, List a )
splitWhen predicate list =
    let
        index =
            findIndex predicate list
    in
        case index of
            Just i ->
                Just (splitAt i list)

            Nothing ->
                Nothing
```

mgold:
How about:

```elm
splitWhen : (a -> Bool) -> List a -> Maybe ( List a, List a )
splitWhen predicate list =
    findIndex predicate list
    |> Maybe.map (\i -> splitAt i list)
```

EverybodyKurts:
Wow, way shorter, lol. Looks good to me. Would love to be able to submit a pull request to this repo (with tests included). You think that'd work?

mgold:
Sure.

(This repo [has no maintainer](https://github.com/elm-community/Manifesto/blob/master/maintainers.md) so I'll go ahead and handle this.)

PR #42, opened 2017-04-25
Create splitWhen function that attempts to split a list at the element where the given predicate is true

EverybodyKurts:
This started as a found here: #41 

Let me know what needs fixed/touched up.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -40,6 +40,7 @@ module List.Extra
         , scanr1
         , unfoldr
         , splitAt
+        , splitWhen
         , takeWhileRight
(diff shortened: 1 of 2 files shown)


mgold:
Looks good!

mgold:
Published as 6.1.0.

BTW, you needed to add `splitWhen` to the `@docs` ist. I took care of it for you since I missed it in review. Strangely, it compiled and tested fine; only `elm package diff` complained.

EverybodyKurts:
Hooray! Thanks for the checking it out and merging it. This is my first contribution to an Elm library.

PR #48, opened 2017-05-30
Add reverseMap

pzp1997:
`reverseMap` is a pretty handy function for efficiently mapping and reversing a `List`. It is inspired by OCaml's [`List.rev_map`](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALrev_map) function.

An example use case is if you have a list of events in chronological order, and you would like to display them in reverse chronological order (like on a résumé or a todo list). Using `reverseMap`, you can convert each event into an `Html msg` and reverse the order of the events all in a single pass.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -9,6 +9,7 @@ module List.Extra
         , maximumBy
         , andMap
         , andThen
+        , reverseMap
         , takeWhile
         , dropWhile
         , unique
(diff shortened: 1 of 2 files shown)


Chadtech:
I really cant see a reason not to merge this. Its just a more performant, and more readable combination of `List.map f >> List.reverse`.

I'll leave this PR open a bit longer, and if nothing changes I'll merge.

PR #43, opened 2017-05-01
Add insertAt for inserting a member at given index

knuton:
Adds the complement to `removeAt`.

PR #55, opened 2017-06-05
Refactor (update|set|remove|swap|split)At for efficiency

### LINKED ISSUE #1: splitAt traverses n elements twice
The current implementation of `splitAt` is a simple `(take n ls, drop n ls)`. This traverses `n` elements twice, first for the take, then for the drop. This could result in bad performance when `n` is a large number.

pzp1997:
I refactored `updateAt`, `setAt`, `removeAt`, `swapAt`, and `splitAt` to make them more efficient. All of the functions are now tail-recursive. In the case of success (i.e. the index is in range), the functions perform two passes over the elements from the start of the list up to and including the given index (and in the case of `swapAt` the larger index). In the case of failure, the functions require only a single pass over the list.

Closes #1 and the second part of #30. I also have a fast implementation for the proposed insertAt function that uses the atHelper, which would close the rest of #30 and #43. I also have another version of the code that changes the return types of `updateAt`, `setAt`, and `swapAt` in accordance with #32 and #36. Just let me know what you need and I can submit additional PRs or add to this one.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -520,11 +520,12 @@ updateIf predicate update list =
 {-| Replace a value at a specific index by calling an update function.
 -}
 updateAt : Int -> (a -> a) -> List a -> Maybe (List a)
-updateAt index update list =
-    if index < 0 || index >= List.length list then
-        Nothing
-    else
-        Just <| updateIfIndex ((==) index) update list
+updateAt index updateFunc list =
+    atHelper index list
+        |> Maybe.map
+            (\( reversedInit, elem, rest ) ->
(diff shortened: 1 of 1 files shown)


pzp1997:
@cbenz Made an interesting point in #36 regarding this PR. Although in theory, these implementations should be faster than the current ones, performance and optimization are a bit tricky in practice. Before we proceed, we should really benchmark the functions so we can truly assess if I was successful. I will try to do this soon. I have also read elm-lang/core#668 about optimizing `List.take`, which has me wondering if we should take a similar approach. I'm going to close this for now, and re-open it once I have the benchmarks.

PR #57, opened 2017-06-11
Fix Travis Builds By Speeding Them Up

prikhi:
Travis builds were failing with the release of elm-test 0.18.5 because they
produced no output for over 10 minutes.

This fixes the issue by using the sample Travis config from the node
test runner's README, which uses sysconfcpus to fix the amount of CPUs
that Elm believes the system has.

This brings Travis builds from 11+ minutes down to 2 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,36 @@
+sudo: false
+
 language: node_js
-node_js:
-  - "5"
+node_js: node
+
+cache:
+  directories:
+    - elm-stuff/build-artifacts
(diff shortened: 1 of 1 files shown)


prikhi:
@Chadtech Thanks! Love it when elm-community test badges are all green :)

Also, please ignore the `long-test-time-osx` branch that I pushed, using it to try & help figure out why elm-test 0.18.4 made our tests go from 2 minutes to 10+ minutes >.<

PR #66, opened 2017-06-23
Optimize findIndices

pzp1997:
[Benchmarks](https://ellie-app.com/3yhz3W6JX96a1/2) show a sizable improvement in runtime over the current implementation. My implementation also generates fewer intermediate data structures, reducing strain on the GC.


Chadtech:
Awesome work btw optimizing all these functions.

PR #71, opened 2017-06-29
Fix documentation

pzp1997:
Fixes some basic formatting mistakes in the documentation. Also, provides a more realistic example (and test) for iterate.


Chadtech:
Thanks for fixing the formatting. Could you add a type signature to `collatz`? 

Im indifferent regarding the new example; I dont see it as better or worse. But I'll defer to you on it unless you feel like explaining it.

pzp1997:
Added the type signature to `collatz`.

Sorry for not explaining the example in the initial PR message. I found the old example to be a poor use of `iterate` because it would be better to use `List.range 2010 2030` instead. The `collatz` example is better because the next value in the sequence depends upon the current value, yet the sequence is not arithmetic or geometric (if it were it would be better to use `List.range` or `List.Extra.initialize`).

PR #72, opened 2017-07-05
adding cycle and zipWithCycle function

mrkaspa:
I added two new functions

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1502,3 +1502,28 @@ greedyGroupsOfWithStep size step xs =
             group :: greedyGroupsOfWithStep size step xs_
         else
             []
+
+
+{-| Repeats the elements of the list until its length
+be the of the size as the first parameter
+    cycle 10 [1, 2, 3]
+      == [1,2,3,1,2,3,1,2,3,1]
+-}
+cycle : Int -> List a -> List a
+cycle len xs =
+    if List.length xs > len then
+        xs
+    else
+        cycle len (xs ++ xs)
+
+
+{-| Zips the first list with the second and repeats the
+elements of the second until it gets the size of the first
+    zipWithCycle [1, 2, 3, 4] [0, -1]
+      == [(1, 0), (2, -1), (3, 0), (4, -1)]
+-}
+zipWithCycle : List a -> List b -> List ( a, b )
+zipWithCycle xs ys =
+    ys
+        |> cycle (List.length xs)
+        |> ListExtra.zip xs
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1518,7 +1518,7 @@ repeat times xs =
 
 {-| Zips the first list with the second and repeats the
 elements of the second until it be greater than the size of the first
-    zipWithCycle [1, 2, 3, 4] [0, -1]
+    greedyZip [1, 2, 3, 4] [0, -1]
       == [(1, 0), (2, -1), (3, 0), (4, -1)]
 -}
 greedyZip : List a -> List b -> List ( a, b )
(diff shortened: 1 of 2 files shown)


Chadtech:
Could you elaborate on a situation in which you would need these functions? I am guessing you have used these functions in your own projects? For what purpose did you use them?

mrkaspa:
@Chadtech yes I needed to mix a list of colors with a list of elements, but there were three colors and a list of n elements so I wanted to repeat the list of colors and in Elm there's no streams or generators to do this

Chadtech:
Sorry I dont understand. You cant mix two different types, and it sounds like you are saying you are mixing something like `Color` and `Html msg`. I am assuming I dont understand you.

And what do you mean by "mix"? Like interspersing?

mrkaspa:
@Chadtech yes it's similar, but interwave doesn't create the tuple, the normal zip do it just until one list runs out of elements, in elixir for example I have this https://hexdocs.pm/elixir/Stream.html#cycle/1 but in elm I can't have lazy collections
 so zipWithCycle makes the second list as big as the first list to complete the zip
```elm
interweave [1,3,5,7,8] [2,4] == [1,2,3,4,5,7,8]
zipWithCycle [1,3,5,7,8] [2,4] == [(1,2), (3,4), (5,2), (7,2), (8,4)]
```

pzp1997:
Perhaps `zipWithCycle` should be called `greedyZip` (in keeping with the naming theme of `greedyGroupsOf`. And of course, this begs the question of whether there should be `greedyMap2`, `greedyMap3`, ... as well.

As for `cycle`, I am not sure if the interpretation of the integer argument is fully intuitive. I would either make it the number of times that the sequence should be repeated, i.e. `cycle 3 [1, 2] == [1, 2, 1, 2, 1, 2]` or *exactly* the length of the return list, i.e. `cycle 4 [1, 2, 3] == [1, 2, 3, 1]`.

mrkaspa:
@pzp1997 You're right, perhaps the names aren't the best, plz check the pr I updated it

Chadtech:
Whats your actual use case tho? What are you using cycle for? Could you talk about the project a bit?

mrkaspa:
@Chadtech I use it here https://github.com/mrkaspa/GitRankWeb/blob/master/src/Views.elm#L150 I use it to mix a tuple of colors with the data of a programming language here you can check it in action http://gitrank.mrkaspa.com/

Chadtech:
Hmm. Okay I see what you are up to.

I wish I was able to think of a better way to do this, but I cant. I would really like to hear other opinions on this. I'll leave this PR open for a while, and maybe look to see if similar functions exist in other programming languages.

Regarding the names, `repeat` is already used in the core list package and does something different, so we should name it something else. Of the names thrown around so far I think the original ones were clearest `cycle` and `zipWithCycle`. Other ideas: `concatSelf`, `cycleZip`, and `zipModulo`.

Chadtech:
In Haskell world theres a function called `cycle` that does largely the same thing. So I suggest we stick with that moving forward.

pzp1997:
I've been thinking about this PR, and here's what I've come up with. I think that the first argument to `cycle` should be the length of the resulting list. This gives people the most flexibility because they can have "partial" cycles if they want, i.e. `cycle 4 [1, 2, 3] == [1, 2, 3, 1]`.

Regarding `zipWithCycle`, I do not think it should be included in the library. Given the version of `cycle` I described above, it is trivial to implement. I also think that its use is sufficiently rare that it's fine if people implement it themselves when they need it. In fact, I think that the use case @mrkaspa linked to above would be better suited by using `cycle` with `map2`.

Chadtech:
@pzp1997  implemented `cycle` so I am closing this PR.

PR #73, opened 2017-07-06
Add count

pzp1997:
`count` is a simple but useful function for checking how many things in a list satisfy some predicate. It is great for when you don't care about the particulars of those things but only their quantity. An example of a good real-world use case is if you are accepting text input from the user and you want to know how many of the words/sentences they used have a particular property.

Suppose we give our students an activity to make complex sentences. A parameter that we might use to check a sentence is seeing if it contains more than one clause, which we can approximate by checking if it contains a `,` character and is adequately long. In order to determine if a student is successful, we simply want to know how many of their sentences satisfy our complexity rule. [We can easily do this using `count`](https://ellie-app.com/3Gm3S9xpXMXa1/0)!

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -67,6 +67,7 @@ module List.Extra
         , elemIndices
         , findIndex
         , findIndices
+        , count
         , zip
         , zip3
         , zip4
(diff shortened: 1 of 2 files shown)


Chadtech:
Cool, that all makes sense to me.

Issue #45, opened 2017-05-09
groupWhile behaves like groupWhileTransitively


ghost:
The docs for `List.Extra.groupWhileTransitively` state that it will "Start a new group each time the comparison test doesn't hold for two adjacent elements".

However, this is the behaviour that `List.Extra.groupWhile` exhibits, which does not seems consistent with its docs.

As far as I can tell the two functions behave identically, and the library provides no way to group together matching elements regardless of their position.

The code
```
module Main exposing (..)

import Html
import List.Extra

main =
    [ 0, 1, 1, 0, 0, 2, 3, 1 ]
        |> List.Extra.groupWhile (==)
        |> toString
        |> Html.text
```
Outputs [[0],[1,1],[0,0],[2],[3],[1]] rather than [[0,0,0],[1,1,1],[2],[3]]

jvoigtlaender:
> As far as I can tell the two functions behave identically

The *documentation* of the package gives examples where the two functions do not behave identically!

ghost:
Sorry.
"Within the context of "Starting a new group each time the comparison test doesn't hold for two adjacent elements", _the two functions behave identically_.

jvoigtlaender:
No. Or, rather: What exactly do you mean? In what sense are the two examples `groupWhile (<) [1,2,3,2,4,1,3,2,1]` and `groupWhileTransitively (<) [1,2,3,2,4,1,3,2,1]`, which demonstrably behave differently, not in that context?

ghost:
The documentation of a function variant states "Start a new group each time the comparison test doesn't hold for two adjacent elements", while the documentation of the original version of the function does not.

Because of this, by reading the documentation, one may be lead to think that the statement above describes a difference in behaviour between the original function and its variant, but this does not seem to be the case.

In other terms: if both functions "Start a new group each time the comparison test doesn't hold for two adjacent elements", why is the statement included only in the variant, rather than in the original?

jvoigtlaender:
Because it DOES NOT HOLD for the original version. Look at the example given for groupWhile with the comparison test (<). There, the test does not hold for the adjacent elements in the third and fourth positions, namely 3 and 2. After all, 3<2 does not hold. And yet, the groupWhile function does not start a new group at that point.

jvoigtlaender:
You may also want to consider that the documentation around there deliberately talks about "equality test" for some things, and "comparison test" for other things.

ghost:
[...]

PR #74, opened 2017-07-13
Minor updates to docstrings, clarification for `groupWhile`

tac-tics:
There was a minor typo calling the test for `groupWhile` an "equivalent relationship". I changed it to "equivalence relation" and wrote out a bit more explicitly for programmers what that means. 

I threw in a few minor grammatical changes.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1254,7 +1254,7 @@ selectSplit xs =
             ( [], x, xs ) :: map (\( lys, y, rys ) -> ( x :: lys, y, rys )) (selectSplit xs)
 
 
-{-| Take 2 lists and return True, if the first list is the prefix of the second list.
+{-| Take two lists and return `True`, if the first list is the prefix of the second list.
 -}
 isPrefixOf : List a -> List a -> Bool
 isPrefixOf prefix xs =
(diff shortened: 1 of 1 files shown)


Chadtech:
What do you mean by "it will give an unexpected answer [ if its not an equivalence relation ]". Its a deterministic process, so the answer follows from the code and the input right? Do you mean the answer is arbitrary or just not useful?

pzp1997:
I interpret it to mean that this behavior is subject to change and should not be relied upon. (Perhaps that phrasing is clearer?) It's also worth noting that there's a somewhat ongoing discussion about getting rid of the non-transitive version of this function in #45.

tac-tics:
@pzp1997 That is essentially what I was getting at.

The main goal of the patch I made was to clean up the explanation for "equivalence relation". But I figured if so much emphasis would be placed on using it in this way (arguably the "correct way"), rather than explain how it works in the non-equivalence relation case, it's best to gently push the user away from doing that. 

Also, just parsing the documentation was causing *me* confusion (knowing full well what an equivalence relation is  :smile:) because it showcased `groupBy (<) [...]` in a code block. Seeing it in a box made it originally seem like "here's an example of how to use it" when in fact that's not how you ever want to use it.

PR #80, opened 2017-09-01
Add cycle

pzp1997:
Related to #72. After some discussion in that PR, it seemed as though the consensus was to make `cycle` accept the length of the resulting list (as opposed to e.g. the number of times the given list should be cycled). This PR contains what I believe to be a relatively fast TCO implementation of `cycle` with the desired behavior (since this would be the initial implementation, there isn't really anything to benchmark it against).

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -29,6 +29,7 @@ module List.Extra
         , filterNot
         , iterate
         , initialize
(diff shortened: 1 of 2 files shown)


Chadtech:
Awesome. Looks merge-worthy to me.

PR #84, opened 2017-10-03
Add cartesianProduct function

rlefevre:
Hi,

I needed this for a project, and this is quite useful, so I thought I might as well do a pull request, in case you think it has its place in `List.Extra`.

I kept the name `sequence` because it is typically named like this for monads (like in `core/Task`) but in the list context, maybe `cartesian` or `cartesianProduct` might be better.

The implementation is almost identical to `sequence` in `core/Task` (but using `lift2` instead of `map2` to explore all combinations).

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -34,6 +34,7 @@ module List.Extra
         , subsequences
         , permutations
         , interweave
+        , cartesianProduct
         , foldl1
         , foldr1
         , indexedFoldl
@@ -92,7 +93,7 @@ module List.Extra
 
 # List transformations
 
-@docs intercalate, transpose, subsequences, permutations, interweave
+@docs intercalate, transpose, subsequences, permutations, interweave, cartesianProduct
 
 
 # Folds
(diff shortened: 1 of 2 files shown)


Chadtech:
Hey this looks pretty good to me.

I find the name `cartesianProduct` a lot nicer. `cartesianProduct` suggest to me exactly what you implemented, while `sequence` doesnt suggest much of anything. That could just be reflecting my unique experience.

rlefevre:
I think you're right. The `sequence` name does not make much sense in a context limited to lists. I renamed it to `cartesianProduct` and made a few syntactic improvements (hopefully).

rlefevre:
Hmm, maybe one issue is that `cartesianProduct [[]] == []` but `cartesianProduct [] == [[]]`.
This seems wrong.

jvoigtlaender:
@rlefevre, mathematically this is exactly correct. What one should expect of a Cartesian product of several lists is something like `length (cartesianProduct lists) == product (map length lists)` where `product = foldr (*) 1`.

The two examples you give do satisfy this property. If you changed those cases, e.g., swapping their results, the property would not anymore hold.

rlefevre:
Ah, thank you very much. I had a doubt about `cartesianProduct [] == [[]]` because ~~`sequence [] == []` in haskell~~. I should have reviewed my maths about the [nullary cartesian product](https://en.wikipedia.org/wiki/Empty_product#Nullary_Cartesian_product) 

> In any category, the product of an empty family is a terminal object of that category [...] and the terminal object is a singleton set.

This was the occasion to add a test. Thanks again.

jvoigtlaender:
I don’t think it is true that in Haskell, the sequence function on the empty list gives the empty list. It certainly gives „return []“, which in the case of the list monad is exactly [[]].

rlefevre:
You're right again, I was tricked by the list monad and `ghci` that produces (as expected) the same output for `[]` and `return []`. Thank you for the clarification.

PR #92, opened 2018-01-09
Fix behavior of foldl1

pzp1997:
The behavior of `foldl1` is not in line with the behavior of `List.foldl`. The problem is that the accumulator for `List.foldl` is the second argument of the "combine" function while for `foldl1`, it is the first argument. The cause of this mistake can be traced most likely to Haskell, where the type of [`foldl`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Prelude.html#v:foldl) is `(b -> a -> b) -> b -> List a -> b`, i.e. the combine function is implicitly flipped.

I have corrected the behavior which incidentally also lead to faster implementations for `foldl1` and `foldr1`. I also wrote better examples for both functions that more clearly demonstrate the difference between them.

*Note that the changes that I made are breaking despite the type of `foldl1` being unchanged.*

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -151,7 +151,7 @@ import Tuple exposing (first, second)
 -}
 last : List a -> Maybe a
 last =
-    foldl1 (flip always)
+    foldl1 always
 
 
 {-| Return all elements of the list except the last one.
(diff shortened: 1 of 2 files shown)


Chadtech:
Great catch @pzp1997 ! Thanks for the PR. Also, could you share a benchmark link?

pzp1997:
Here is the [`foldl1` benchmark](https://ellie-app.com/f59bYsXtFa1/1) and the [`foldr1` benchmark](https://ellie-app.com/sDtg53GXsa1/0).

Chadtech:
The benchmarks look good to me! Thanks for the contribution @pzp1997 . Since this is a breaking change, I will merge this on our next major release.

I wonder if its worth updating the future docs to mention this breaking change? Heres a PR I just made to include an explanation of the current behavior in our current docs #94

PR #93, opened 2018-01-11
Fix performance regression in List.Extra#init

brianvanburken:
Commit 5fa63ca introduced a performance improvement for non empty lists
but also a ~70% performance regression for empty lists. This commit
fixes the regression while also improving the performance ~30% for
empty lists before 5fa63ca.

https://ellie-app.com/5K7f2nGwsa1/1 for benchmarks.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -170,10 +170,15 @@ last items =
 -}
 init : List a -> Maybe (List a)
 init items =
-    items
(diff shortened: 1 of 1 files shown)


zwilias:
Good stuff! :)

Chadtech:
Nice! Looks good to me. Ill merge this soon.

PR #96, opened 2018-03-17
Add cartesianProduct2

mgold:
This is a variation of `cartesianProduct` for when exactly two lists are being used. Instead of a list of lists -- the inner lists having length 2 -- the inner lists are replaced by pairs.

This came up in a project I was working on. Just as a sketch of utility, expanding a tree according to all possible transformations of an element:

```elm

type Tree a = Leaf a | Branch (Tree a) (Tree a)

transform : (a -> List a) -> Tree a -> List (Tree a)
transform f tree =
  case tree of
     Leaf x -> f x |> List.map Leaf
     Branch left right ->
        let
           lefts = transform f left
           rights = transform f right
        in case ( lefts, rights ) of
           ( [], [] ) -> [ tree ]

           ( _, [] ) ->
                    List.map (\bc -> Branch bc right) lefts

           ( [], _ ) ->
                    List.map (\bc -> Branch left bc) rights

           ( _, _ ) ->
                    cartesianProduct2 lefts rights
                        |> List.map (uncurry Branch)
```

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -36,6 +36,7 @@ module List.Extra
         , permutations
         , interweave
         , cartesianProduct
+        , cartesianProduct2
         , foldl1
         , foldr1
         , indexedFoldl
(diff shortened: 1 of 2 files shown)


pzp1997:
I'm pretty sure that `cartesianProduct2` is just `List.Extra.lift2 (,)`. Please correct me if I'm wrong. I think that having such a function might be encouraging an anti-pattern since typically you want to do something with the pairs for which `lift2` is more suitable.

mgold:
Yes, that's correct. Thank you for pointing that out.

PR #97, opened 2018-04-15
added removeSubsequentDuplicates to Extra.elm

jackhp95:
This is my first submission to OSS!

I ran into a situation where I needed a function like this, and I couldn't find a function that would remove subsequent duplicates in this library or in the elm-core. So I made removeSubsequentDuplicates and I figured someone may find it useful. 

I hope I handled things correctly, please let me know of things I could improve. Thank you!

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -84,6 +84,7 @@ module List.Extra
         , greedyGroupsOf
         , greedyGroupsOfWithStep
         , groupsOfVarying
+        , removeSubsequentDuplicates
         )
 
 {-| Convenience functions for working with List
@@ -91,7 +92,7 @@ module List.Extra
 
 # Basics
 
-@docs last, init, getAt, (!!), uncons, maximumBy, minimumBy, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, replaceIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
+@docs last, init, getAt, (!!), uncons, maximumBy, minimumBy, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, replaceIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, removeSubsequentDuplicates, filterNot, swapAt, stableSortWith
 
 
 # List transformations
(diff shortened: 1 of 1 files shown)


pzp1997:
This seems like a good function to have, but I'm not crazy about the name. You should also write some tests for your function in [list-extra/tests/Tests.elm](https://github.com/elm-community/list-extra/blob/master/tests/Tests.elm).

I haven't done any benchmarking, but I have a feeling that the following `foldr` implementation is faster.

```
removeSubsequentDuplicatesHelper : a -> List a -> List a
removeSubsequentDuplicatesHelper x acc =
    case head acc of
        Just y ->
            if x == y then
                acc
            else
                x :: acc

        Nothing ->
            x :: acc


removeSubsequentDuplicates : List a -> List a
removeSubsequentDuplicates list =
    foldr removeSubsequentDuplicatesHelper [] list
```

Here's another possible implementation that is potentially even faster than the foldr implementation but not stack safe.

```
removeSubsequentDuplicates : List a -> List a
removeSubsequentDuplicates list =
    case list of
        x :: y :: ys ->
            if x == y then
                removeSubsequentDuplicates (y :: ys)
            else
                x :: removeSubsequentDuplicates (y :: ys)

        _ ->
            list
```

I think we should do some benchmarking to quantify how fast each option is before making a decision on which to use.

jackhp95:
Ah! Thanks for the quick feedback! I'll work on benchmarking and tests and I'll get back to you. :)

Chadtech:
Thanks @jackhp95 and @pzp1997 !

Yeah I also am not crazy about the name. Reading it left me with the impression that all subsequent duplicates in the list were filtered, not just the immediate next elements. Maybe `uniqueNeighbors`? Just an idea.

And then, echoing what @pzp1997 said, if you could get the code working, add a test to our test file, and maybe benchmark a few alternative implementations to see which one is fastest. For bench marking you can use Ellie app. I always just copy and paste the code of an old bench mark we did for `List.Extra` and start from there, such as this one from another PR: https://ellie-app.com/5K7f2nGwsa1/1

Thanks for the submission! Congratulations on contributing to OSS. Let me know if you need any help.

pzp1997:
@jackhp95 Sorry, if I came off as rude earlier. I hope I didn't discourage you from contributing! Submitting PRs is a very rewarding process, and we are here to help you get your first one accepted. Please reach out with any questions.

Anyway, here's one more implementation for benchmarking. I expect it to be comparable to the other `foldr` implementation. It might even be slightly faster since the helper does not call `head` for each element in the list.

```
help : a -> ( a, List a ) -> ( a, List a )
help y ( x, xs ) =
    if x == y then
        ( x, xs )
    else
        ( y, x :: xs )


removeSubsequentDuplicates : List a -> List a
removeSubsequentDuplicates list =
    case list of
        [] ->
            []

        x :: xs ->
            uncurry (::) (foldr help ( x, [] ) xs)
```

PR #101, opened 2018-05-06
Define & test unconsLast

EverybodyKurts:
Handy for when you want to remove the last element of a list.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -306,10 +306,14 @@ uncons xs =
 
 -}
 unconsLast : List a -> Maybe ( a, List a )
-unconsLast =
-    List.reverse
-        >> uncons
-        >> Maybe.andThen (\( lastElem, reversedList ) -> Just ( lastElem, (List.reverse reversedList) ))
+unconsLast list =
+    case List.reverse list of
+        [] ->
+            Nothing
+            
+        last :: rest ->
+            (last, List.reverse rest)
+                |> Just
 
 
 {-| Find the first maximum element in a list using a comparable transformation
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -5,6 +5,7 @@ module List.Extra
         , getAt
         , (!!)
         , uncons
+        , unconsLast
         , minimumBy
         , maximumBy
         , andMap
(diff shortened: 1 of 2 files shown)


Chadtech:
Hey @KurtRMueller , thanks for contributing.

Do you have a use case for this function in mind? It sounds like a pretty basic function, but I always want some practical context before merging stuff.

EverybodyKurts:
I've been using it when I want to dig through a list, split it apart, and then combine the first and last elements of the two lists.

I've also been using it to get an event's first and most recent date times.

I can try to come up with a more specific examples if you like. Let me know.

Chadtech:
Could you elaborate a little bit on your first example? Both of your cases seem like they could be accomplished with `List.Extra.last`, so I am trying to figure out whats _uniquely_ valuable about `unconsLast`.

I split things into `first :: rest` all the time, and I suspect a lot of people do, so it doesnt seem implausible to me that someone would need `(rest, last)`, I just dont have a clear picture right now.

EverybodyKurts:
Similar to the value that `uncons` provides, in which you get the first element of the list and the rest of the list with the first element removed, `unconsLast` allows you to get the last element of a list along with the rest of the list with the last element removed. 

As for the first example, I've been working on app that allows the user to create a passage of text where each individual letter is a `Character`. The app allows `Characters` to be highlighted and combined into `Word`s which is composed of a definition and a list of `Characters`. If a user highlights a character and then sweeps their mouse backwards, the app has been finding where the list of characters needs to split, divides the original list into two parts, and then has been using `unconsLast` to iteratively take away from the first part of the list, add to the newly created `Word`, and keep the remaining `Character`s of the first part of the list.

I hope the above blurb made sense. The value in `unconsLast` is similar to the  value of `uncons` where you can get both the first and remaining elements of the list. `unconsLast allows me to get both the last element of the list and the remaining elements of the list.

Chadtech:
I think I understand what use case you are describing, but I dont know why you would have to iterate from the end of the list of `Character` to accomplish that.

Either way, lets merge this.

Could you fix the errors on this PR, and also I think I found a more performant implementation of `unconsLast`. You can see it with a benchmark here: https://ellie-app.com/t6KfjbzNnWa1 Would you mind using this implementation instead?

EverybodyKurts:
That was a nice little demo. I never knew about the Benchmark library. I can use that implementation instead.

EverybodyKurts:
Hey @Chadtech, everything should be good to go. Let me know if anything is needed.

Chadtech:
Looks good! Thanks for contributing

Issue #49, opened 2017-06-01
groupWhile should return a List of Nonempty, not a List of List


ryan-senn:
`groupWhile : (a -> a -> Bool) -> List a -> List (List a)`

Should be 

`groupWhile : (a -> a -> Bool) -> List a -> List (Nonempty a)`

The issue is that Nonempty is a different package obviously. But as it stands it's kind of wrong, no? Maybe it belongs on the other package though.

ckoster22:
If `groupWhile` was given an empty list wouldn't it produce an empty list?

Chadtech:
`groupWhile f []` would be empty. So its possible.

jvoigtlaender:
But is `groupWhile` the only function in this package here that would be more accurately typed by replacing some `List` by a `Nonempty`? I don't think so. Just going through the package documentation page from top to bottom, I quickly encountered another one: `setAt` can never return `Just []`, so its most accurate return type would be `Maybe (Nonempty a)`, not `Maybe (List a)`. 

I'm not saying that the type of `setAt` should be changed. But I find it strange to somewhat arbitrarily pick one function from the package and bring `Nonempty` in for it. Wouldn't a more meaningful approach be to:
1. go trough all functions in the package and check for which ones such a change would be possible,
2. decide whether it's worth it overall.

pzp1997:
@jvoigtlaender The return type for `setAt` recently changed to get rid of the `Maybe`, so it no longer makes sense to return a `Nonempty` (the same applies to most/all of the other *At methods).  If you pass it an index that is less than 0 or greater than or equal to the list's length, it will not insert the element and is essentially a no-op. Therefore, when given an empty list, it will always return an empty list. Eliminating the methods whose return type recently changed, I don't see any obvious candidates that would benefit from `Nonempty`.

jvoigtlaender:
Okay, fine for `*At`, but it was really just the first example I came upon. My real question is not about `setAt`, but about whether, if `Nonempty` is brought in as a dependency for this package, more thought should go into what that means (or could mean) for the whole set of functions, not just the individual function `groupWhile`.

z5h:
@Chadtech `groupWhile f []` would still return `List NonEmptyList`. `[]` *is* a list of non-empty lists. (In fact, it's a list of ANYTHING.)

Chadtech:
Okay I see now @z5h . Yeah I was wrong.

z5h:
[...]

PR #103, opened 2018-05-31
Change group/groupWhile to use a tuple to represent a non-empty list

loganmac:
Quote from @evancz in the slack: 

> Someone mentioned that using `List.Extra.groupWhile` is annoying because you are guaranteed to get at least one entry in every single list, but you then have to pattern match and handle the empty case anyway. Can it be changed to the following type?
> 
> ```groupWhile : (a -> a -> Bool) -> List a -> List (a, List a)```
>
> I am going to be emphasizing to people that “non-empty lists” are not hard to represent, no special library needed. I think doing it as a tuple is a fine way

This PR does just that.

### DIFF
--- a/elm-package.json
+++ b/elm-package.json
@@ -1,5 +1,5 @@
 {
-    "version": "7.1.0",
+    "version": "8.0.0",
     "summary": "Convenience functions for working with List",
     "repository": "https://github.com/elm-community/list-extra.git",
     "license": "MIT",
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1277,7 +1277,7 @@ stripPrefix prefix xs =
     group [1,2,2,3,3,3,2,2,1] == [[1],[2,2],[3,3,3],[2,2],[1]]
 
 -}
-group : List a -> List (List a)
+group : List a -> List (a, List a)
 group =
     groupWhile (==)
 
@@ -1297,7 +1297,7 @@ For non-equivalent relations `groupWhile` has non-intuitive behavior. For exampl
 For grouping elements with a comparison test which is merely transitive, such as `(<)` or `(<=)`, see `groupWhileTransitively`.
 
 -}
-groupWhile : (a -> a -> Bool) -> List a -> List (List a)
+groupWhile : (a -> a -> Bool) -> List a -> List (a, List a)
 groupWhile eq xs_ =
     case xs_ of
         [] ->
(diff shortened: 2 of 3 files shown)


pzp1997:
See #49 for related discussion.

Chadtech:
Discussion seems to have fizzled out on this. To move forward, how about we merge this, in conjunction with a handful of non empty list helper functions, such as..
```elm
mapNonEmpty : (a -> b) -> (a, List a) -> (b, List b)
toNonEmpty : List a -> Maybe (a, List a)
fromNonEmpty : (a, List a) -> List a
```

..is that a good plan?

pzp1997:
I'm kind of against that proposal. At the end of the day we'll have a whole lot of duplicate functions just for dealing with non-empty lists that will pollute the library (you mentioned 3 or 4 functions in particular but why should we stop there?). Maybe this stuff belongs in a different package or module. Also, FYI toNonEmpty is the same thing as uncons.

pzp1997:
If we do decide to merge this PR, which again I don't think we should do, some work needs to be done. 

- There is one other grouping function that should be updated for consistency, groupWhileTransitively.
- There's a minor formatting issue on line 1311.
- The version probably should not be bumped in a pull request.

Chadtech:
We talked about this in the issues and came to the conclusion that we should merge this without any helpers. If anyone needs any helpers later we can always add them.

Chadtech:
Oh, and I will publish this with the next major update, which I would estimate will be within a month or two.

pzp1997:
Please make sure the comments I made above are addressed before publishing.

Chadtech:
Right, forgot about that, sorry. I will write that down and make sure those items are in the next release. I am going to have to sit down and solve some merge conflicts for the next release anyway. Unless @loganmac wants to do it (no pressure tho).

Chadtech:
Okay, I have implemented those remaining changes in the `0.19` branch.

Rather than update `groupWhileTransitively` I just deleted it. I went to update it, but that involved figuring out what it did exactly compared to `groupWhile`. Then I found #45 , and realized the distinction was in dispute. I agreed with the opinions expressed in that issue we should just delete `groupWhileTransitively`, and so thats what I did. 

Here is a demo: https://ellie-app.com/32KL9L2NtPKa1

Im worried this is too rushed. Its just some code I wrote up and pushed. So if you have any review of it I  would appreciate it.

pzp1997:
LGTM. Sorry for not getting to it sooner.

PR #108, opened 2018-08-27
Update tests for 0.19

rlefevre:
I needed this to validate some new tests, so I thought I could as well make a pull request.

I tested with elm-test@0.19.0-beta4, there was an issue with `groupWhile` that is fixed by the second commit (it seems that there was a mistake while updating the tests during last `groupWhile` behavior change).


Chadtech:
Okay. These tests work for me when I run them locally with the latest version of elm-test (0.19.0-beta5). 

Thanks a lot for doing this, its something that needed to get done.

rlefevre:
By the way, I noticed too late that I forgot to remove `tests/.gitignore`. You may want to remove it as I don't think that it will be useful anymore.

Chadtech:
Great. Yeah and the travis set up needs to be updated too, but I think I will update that when elm-test is past beta.

Issue #112, opened 2018-09-03
Upgrading tests


Chadtech:
Elm 0.19 is out, and out tests stuff is all out of wack. Heres what needs to get done:
- [x] Upgrade the tests file for 0.19 (Done, thanks to @rlefevre )
- [x] Remove `tests/.gitignore`
- [x] Fix travis set up

We can probably wait until elm-test is past beta for fixing the travis stuff.

Chadtech:
Okay I have all this taken care of in #115

Chadtech:
I am going to merge #115 soon. Firstly because a few people have noticed and made PRs for documentation problems that can be resolved by this topic, and secondly because Im thinking now that maybe we cant wait on elm-test moving beyond beta (this happened when we waited on Elm 0.19 too, so maybe I should learn not to wait on dependencies).

Issue #111, opened 2018-09-03
`group` example in docs returns wrong type


Janiczek:
`group` returns `List (a, List a)` but the example doesn't reflect that:

```elm
group  [1,2,2,3,3,3,2,2,1] == [[1],[2,2],[3,3,3],[2,2],[1]]
```

Chadtech:
Thanks! I didnt notice this. Yeah in `8.0.0` the group functions changed pretty fundamentally. I'll fix this soon.

pzp1997:
I think we should start using [elm-verify-examples](https://github.com/stoeffel/elm-verify-examples) so that this will not happen in the future. See issue #50.

Chadtech:
Sounds good to me. Ilias recommended that too. Lets do it.

Chadtech:
We are using elm-verify-examples now, and this particular issue was resolved.

PR #113, opened 2018-09-21
fixed incorrect example of using groupWhile

nikolakasev:
While refactoring an application of mine I saw that groupWhile has changed. Running the example gives a misleading result.


pzp1997:
I think we should batch this with #111 and #112.

Chadtech:
Thanks @nikolakasev

Issue #105, opened 2018-07-25
Non-Empty Discussion


Chadtech:
We recently merged a change to `groupWhile` to make it deal with non-empty lists `(a, List a)`. Until now, we have kind of steered clear of non-empty lists.

So now we have a little bit of non-empty lists in List.Extra. Is that a good move? A bad move? Where do we draw the line? Do we need to draw any lines? Bottom line question is, should we have non-empty list support in List Extra?

My opinion is that theres no problem with a some non-empty list functions. We dont have to support every possible operation, but a few helpers that can make it easy to transform non-empty lists to lists and vice versa would be nice.

robinheghan:
I'm not a huge fan myself, as I usually pass the result of `list-extra` fns to other functions which deal with a `List`. If `list-extra` return `(a, List a)` everywhere, I'll often have to convert to from that to a regular `List`, which already composes well.

It's not a big deal, but personally I don't often have use of non-empty lists.

pzp1997:
In my own code, I often will use `uncurry (::)` to convert `( a, List a ) -> List a`. I think it works pretty well when pipelined. @Skinney, do you think adding a function to that effect to list-extra would be helpful?

```
fromNonempty : ( a, List a ) -> List a
fromNonempty ( x, xs ) =
    x :: xs
```

robinheghan:
I think that would be a good idea. I also think it would be good to use a type alias in those functions which return a nonempty list, to make it more clear :)

```elm
type alias Nonempty = (a, List a)
```

Chadtech:
I also like `fromNonempty`.

The alias, I am not so sure. I see `Nonempty` as _adding_ information about the use of the data, but also _removing_ information about the underlying data structure. For example, you cant immediately tell that `Tuple.first` can be used on `Nonempty a` but you could from seeing `(a, List a)`. This might disadvantage people newly approaching codebases that use the `Nonempty` alias.

I could be wrong. Maybe this is a worthy trade off. This is the kind of thing I wish we could obtain empirical data on, but I suppose we never will.

janwirth:
I once wrote this piece of code where the non-empty actually came in quite useful!
I found it often improves the way the my code evolves - similar to the constraits of elm-lang put upon you: When it starts feeling clunky you are doing something wrong.
```elm
-- GROUPING


type alias Groups a =
    List (Group a)


type alias Group a =
    { name : String, items : List a }


type alias Config a =
    { outerOrder : Comparison a
[...]

PR #114, opened 2018-09-26
Add gatherEquals function.

robinheghan:
This is similar to `group`, except it groups all equal elements in a single sub-list. I'm open to change the name to something better.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -112,7 +112,7 @@ module List.Extra
 
 # Sublists
 
-@docs splitAt, splitWhen, takeWhileRight, dropWhileRight, span, break, stripPrefix, group, groupWhile, inits, tails, select, selectSplit, gatherEquals
+@docs splitAt, splitWhen, takeWhileRight, dropWhileRight, span, break, stripPrefix, group, groupWhile, inits, tails, select, selectSplit, gatherEquals, gatherEqualsBy, gatherWith
 
 
 # Predicates
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1679,9 +1679,9 @@ greedyGroupsOfWithStep size step xs =
 {-| Group equal elements together. This is different from `group` as each sublist
 will contain *all* equal elements of the original list.
 
-    gatherEquals [1,2,1,3,2] == [[1,1],[2,2],[3]]
+    gatherEquals [1,2,1,3,2] == [(1,[1]),(2,[2]),(3,[])]
 -}
-gatherEquals : List a -> List (List a)
+gatherEquals : List a -> List (a, List a)
 gatherEquals list =
     gatherEqualsWith (==) list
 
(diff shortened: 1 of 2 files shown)


pzp1997:
I really like the idea of this function but I'm wondering how useful this is without allowing for a custom equality function. It seems to me that the only useful thing you can do with the output is to calculate how many copies of each element there are. (Please tell me if I am overlooking something here.)

In that case maybe `List (a, Int)` would be a better return type for `gatherEquals`? We could also add a variant `gatherEqualsWith : (a -> a -> Bool) -> List a -> List (a, List a)` where the `(a, List a)` is to show that the sublists are non-empty.

robinheghan:
We actually use it in production ^^

It's both for counting the number of things, but we also give each element an index and render each section together. So it does have more uses then just counting them.

Your arguments are valid, so I'm open to adding both `gatherEqualsWith` and `gatherEqualsBy` :)

Chadtech:
How exactly are you using it?

I think I can kind of see how you might be using it. Something like `List.repeat 5 xView` would work, but you already have a list `[ x, x, x, x, x ]` so its easier to render all the items in the list, rather than count them and render that number of xViews. Like, an adventure video game where you have an inventory, and you might have 10 potions that are identical data with identical views. It would make more sense to just render all 10 potions since thats the data available to you then it would be to count the potions and do `List.repeat potionCount potionView`

Is that right? Am I anywhere close?

I also agree about adding `gatherEqualsWith` and `gatherEqualsBy`. Seems like a good idea to me.

robinheghan:
Added `gatherEqualsBy` and `gatherEqualsWith`.

The way we use it is that we get a list of purchased tickets from our backend. We want to summarize what you've bought (eg. 2 adult tickets, 1 child ticket), and if the user clicks "view details" we want to print a human readable description of each ticket as well.

So while @pzp1997 would work, we would likely need to use `List.repeat` to render the detailed view later, or just render the backend provided list.

For us, grouping the elements then counting and rendering seemed like the better option.

pzp1997:
I can live with your desire to avoid `List.repeat` in the view function. How do we feel about making the return type `List (a, List a)` to reflect that the groups are guaranteed to be non-empty? (See #49 and #105 for related discussions about the place of non-empty in List.Extra.)

robinheghan:
Updated to return (a, List a)

pzp1997:
I have a backlog of work right now but I can review your updated code in a couple days if no one gets to it first.

Chadtech:
The code looks good to me. I havent run the tests yet to be 100% certain.

I do notice two things in the documentation
1. The `gatherEqualsBy` example uses `.name` but all the records contain `{ age = .. }`.
2. The `gatherEqualsWith` example uses the function `gatherEquals`, but is otherwise correct.

Also, its just occurring to me that the names `gatherEqualsBy` and `gatherEqualsWith` could be confusing. For example, you could write a comparison function that returns `True` if two `Int` are not more different than 1, grouping `1` and `2` together, even tho `1` and `2` arent equal. What do you all think about `gatherBy` and `gatherWith`?

robinheghan:
I agree with `gatherWith`. However, `gatherEqualsBy` should remain the same, as it provides an equality check on whatever the function returns, it doesn't allow the user to implement custom equality checking.

pzp1997:
I think you need to add `gatherEqualsBy` and `gatherWith` to the \@docs list. Also might be worth adding one sentence about the order in which the groups are returned and the order of the elements within each group. Other than that LGTM.

Chadtech:
Great. Looks good to me too. Thanks @Skinney !

PR #116, opened 2018-10-08
Fix groupWhile example

Arkham:
Hi there! Just realized there is a small typo in the doc for groupWhile :)


Arkham:
Aha, just saw you added elm-verify in another branch, so this isn't needed anymore!

Chadtech:
Thanks for the PR anyway. Perhaps I should merge that soon to prevent this from happening again.

PR #117, opened 2018-10-20
Tests for maximum by and minimum by

frankschmitt:
I noticed that there were no tests for maximumBy and minimumBy, so I added some simple tests for both of them (in essence, I copied the tests for maximumWith / minimumWith and changed the predicate)


Chadtech:
Thank you!

PR #118, opened 2018-12-03
Add `pairings` function

Janiczek:
```elm
pairings [1,2,3,4]
--> [ (1,2), (1,3), (1,4), (2,3), (2,4), (3,4) ]
```

This is essentially doing the equivalent of the following Python loop:
```python
def pairings(xs):
    return [(xs[i], xs[j]) for i in range(len(xs))
                           for j in range(i+1, len(xs))]
```
It's useful eg. for Advent of Code puzzles, or any occasion where you need to check elements of a list against each other, without duplicates and without checking against self (`(x,x)`).

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1071,13 +1071,7 @@ cartesianProduct ll =
 (Essentially, enumerate the possible "handshakes.")
 
     pairings [ 1, 2, 3, 4 ]
-    --> [ ( 1, 2 )
-        , ( 1, 3 )
-        , ( 1, 4 )
-        , ( 2, 3 )
-        , ( 2, 4 )
-        , ( 3, 4 )
-        ]
+    --> [ ( 1, 2 ), ( 1, 3 ), ( 1, 4 ), ( 2, 3 ), ( 2, 4 ), ( 3, 4 ) ]
 
 In this example, everybody shakes hands with three other people.
 
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1,7 +1,7 @@
 
 module List.Extra exposing
     ( last, init, getAt, uncons, unconsLast, maximumBy, maximumWith, minimumBy, minimumWith, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, setIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
-    , intercalate, transpose, subsequences, permutations, interweave, cartesianProduct
+    , intercalate, transpose, subsequences, permutations, interweave, cartesianProduct, pairings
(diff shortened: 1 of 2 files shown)


Chadtech:
Hey @Janiczek !

This looks good to me. A few points tho:
1. Im guessing you were working on the advent of code challenge, would you mind linking to the problem and your solution that used this function?
1. Would you mind adding a few tests to our test file?
1. Maybe the name could be more specific? Im just worried that this function will get mixed up with other functions that involve lists and pairing and that it seem like "pairings" describes both equally well. Like "zip" is a kind of "pairings" it seems. Some alternative name ideas: `pairOwnMembers`, `everyPairOf`

prikhi:
> Maybe the name could be more specific? Im just worried that this function will get mixed up with other functions that involve lists and pairing and that it seem like "pairings" describes both equally well. Like "zip" is a kind of "pairings" it seems. Some alternative name ideas: pairOwnMembers, everyPairOf

It's called the [Cartesian Product](https://en.wikipedia.org/wiki/Cartesian_product), but that might too mathy.

Maybe `pairPermutations`?

Edit: Nevermind, I would assumed the permutations would include `(2,1)`, not just `(1,2)`.

Janiczek:
@Chadtech 
1. Yes, it was [AoC 2018, problem 3](https://adventofcode.com/2018/day/3). The `pairings` function allows me to simplify the code [in this way](https://github.com/Janiczek/advent-of-code/commit/0b251696d0b0e06070485b3ddd42ae36607a582d).
2. Of course, yes I'll add them. In a moment :)
3. I guess the mathematical name is "combinations without repetition". The usual example is "handshakes". I'm partial to changing the name to `combinations`. What do you think?

Janiczek:
OTOH, this is not the general function for combinations -- it only computes 2-combinations.

Chadtech:
Yeah seems like `combinations` is more general. Im worried that even `pairings` is too general. How would you know from the name that it doesnt include `(1,2)` and `(2,1)` in the output list?

Seems like the most specific name would be `handshakes`. What do you think of that? Im worried about the name and the documentation being too mathematical and people not knowing what a "handshake" is (I didnt, I had to look it up, and while it was straight forward reading about it ideally looking up the term wouldnt be a necessary step).

Chadtech:
`pairsWithoutReptition`?
`uniquePairs`?

pzp1997:
FWIW I am a fan of `handshakes`.

Janiczek:
@Chadtech your `uniquePairs` sound the best to me.

The bit about order not mattering (`(a,b)` being the same as `(b,a)`) would have to be deduced from the documentation and/or the examples, but I think there's only so much meaning you can put into a function name...

Chadtech:
Okay. Lets do `uniquePairs` then? With that change, I'd be happy to merge.

Chadtech:
Perfect! Thanks for your PR

Issue #75, opened 2017-07-18
Group methods will crash if used with a large list


andys8:
Several methods are implemented in a way that is not tail recursive. I'm not sure what's the current state in elm and in js regarding tail call optimization (see https://lisplog.org/elm_converts_tail_calls_to_loops.html) or if [elm-lang/trampoline](http://package.elm-lang.org/packages/elm-lang/trampoline/1.0.1/Trampoline) could help. Open for discussion :)

#### Reproduce issue

```elm
list : List (List Int)
list =
    List.range 0 100000
    |> List.Extra.group
```
![image](https://user-images.githubusercontent.com/13085980/28313127-f78993cc-6bb5-11e7-9e29-965b4614ac5e.png)

https://ellie-app.com/3MxzbGDZdNSa1/1

#### Related issues
#61

pzp1997:
This is an interesting discussion to have. Ultimately, the decision comes down to whether we want to prioritize making the implementations 100% safe or if we want them to be as fast as possible.

On the one hand, having a guarantee that this library will never cause your app to crash is a strong statement, particularly in Elm. It is also very unlikely that whatever performance degradation this may cause would have any noticeable impact on real-world apps due to overwhelming cost of DOM manipulations.

On the other hand, most non-TCO functions will only break for lists of more than 5k-10k elements, which is quite a lot. I can't remember ever needing a list that large in Elm (although I'm sure other folks have their use-cases). An argument could be made that if you are using lists that are that large then you should consider using a more appropriate data structure.

Note that there is a middle-ground. It is possible to start off with the faster, non-TCO implementation and then switch to the TCO implementation after processing X elements (for some constant X around 3k). This is how `take` (see elm-lang/core#668) and likely the v0.19 `foldr` (see elm-lang/core#872) are implemented in core. I tried this approach for a few functions in this library, like `updateAt` and `removeAt`, but they actually performed worse than the current implementations :man_shrugging:.

andys8:
I think the elm ecosystem should prefer working and stable implementations. If this library is not meant to be very stable there should be a warning and it probably can't be used in production.

This commit is interesting. It solves the same potential problem in `elm-lang/core`.

https://github.com/elm-lang/core/commit/decbc93025ba38ee9c5613d311e5b979452dfcd3

pzp1997:
Personally, I agree with you that everything should be stack-safe, but I do not think it is as simple a decision as you are making it out to be.

[...]

PR #119, opened 2019-02-21
rewrite groupWhile to be tail-recursive

BrianHicks:
I have a very long list for which groupWhile blows the stack! Oh no! It looks like the accumulator function itself was tail-recursive, but not one of the functions it called. This unifies them and adds a bunch of comments to explain what's going on for the next person who needs to look at this code.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1592,7 +1592,7 @@ groupWhileTailRecursive isSameGroup remaining acc =
         -- we're well underway in this case; we have *both* a new item to look
         -- at and a existing item to compare it to. We'll finally use the
         -- grouping function to figure out if we need to split into a new group!
-        ( now :: next, ( previous, group ) :: finished ) ->
+        ( now :: next, ( previous, alreadyGrouped ) :: finished ) ->
             if isSameGroup now previous then
                 -- note that we're replacing the first item of this tuple here,
                 -- and storing the previous item at the front of the list. That
@@ -1600,7 +1600,7 @@ groupWhileTailRecursive isSameGroup remaining acc =
                 -- reverse 'em when we finish the group.
                 groupWhileTailRecursive isSameGroup
                     next
-                    (( now, previous :: group )
+                    (( now, previous :: alreadyGrouped )
                         :: finished
                     )
 
@@ -1608,7 +1608,7 @@ groupWhileTailRecursive isSameGroup remaining acc =
                 groupWhileTailRecursive isSameGroup
                     next
                     (( now, [] )
-                        :: reverseNonEmpty ( previous, group )
+                        :: reverseNonEmpty ( previous, alreadyGrouped )
                         :: finished
                     )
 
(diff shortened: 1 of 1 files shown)


BrianHicks:
bahh, sorry for the failures here. I've reimplemented this in a private repo to avoid the problem and this is more an extraction than a really planned contribution. I'll fix the tests and update here when it's ready to look at.

BrianHicks:
alright! Looks like this passed, ready for review!

pzp1997:
So it looks like the day when someone has a large enough list in practice to break the group functions has finally come.

> On the other hand, most non-TCO functions will only break for lists of more than 5k-10k elements, which is quite a lot. I can't remember ever needing a list that large in Elm (although I'm sure other folks have their use-cases).

\- me, https://github.com/elm-community/list-extra/issues/75#issuecomment-316088047.

Perhaps we should reconsider the discussion about whether to prioritize stack safety or performance in list-extra. In light of the new evidence presented by Brian, I'm leaning towards the former.

BrianHicks:
I think that may be a false dichotomy in this case. This code makes exactly two passes over the list as a whole, and two passes over every group… so I guess it's `O(4n)`. I don't see anything that's an obvious performance improvement here, regardless of stack safety. If you see something, I'm happy to add it!

BrianHicks:
mmm, sorry, that should have been "one pass over the list, one pass to reverse every group, and one pass to reverse the entire list" so I guess it's actually `O(3n)`

pzp1997:
> I don't see anything that's an obvious performance improvement here, regardless of stack safety.

Yeah, my original comment was made when the group functions looked a bit different. A lot of things changed with 5d30fa7 making the current implementation neither performant as it could be nor tail recursive.

---

I think the following stack safe implementation using `List.foldr` is simpler to read and might even be faster, particularly on lists with fewer than 500 elements due to `foldr`'s "switching behavior" between non-TCO and TCO. I don't have time to run benchmarks unfortunately.

```
groupWhile : (a -> a -> Bool) -> List a -> List ( a, List a )
groupWhile isSameGroup items =
    List.foldr
        (\x acc ->
            case acc of
                [] ->
                    [ ( x, [] ) ]

                ( y, restOfGroup ) :: groups ->
                    if isSameGroup x y then
                        ( x, y :: restOfGroup ) :: groups

                    else
                        ( x, [] ) :: acc
        )
        []
        items
```

Here's an Ellie with this implementation and the relevant tests from list-extra: https://ellie-app.com/4N7MXvP5Bpka1.

BrianHicks:
yeah, I think that might also be faster because it doesn't have to reverse the individual groups. If I understand correctly then:

- one scan to get to the end of the list
- one scan to go back to the head, accumulating items the whole time

`O(2n)`.

pzp1997:
LGTM!

Chadtech:
Awesome! This looks good to me too.

Thanks @BrianHicks for the PR, and the code comments. I kind of wish we could get some of those comments into this PR with this new implementation. People root around in the source of List.Extra; I think it would get some visibility.

I would like to benchmark the new implementation, just to characterize it and make sure there are no surprises. Unfortunately, Ellie is not working right now for me, how about I bench mark this function sometime soon, like in the next day or so? And then, we merge this PR?

BrianHicks:
Sounds great! No rush on my part; we’ve fixed things locally.

Chadtech:
Okay! I bench marked it. You can see it here: https://github.com/Chadtech/elm-groupwhile-benchmark-0 (couldnt do Ellie for some reason, so I just made a repo).

Results are that this new implementation is about 15% slower. Not too bad I think.

PR #126, opened 2019-08-18
Simpler implementation of the "cycle" function

flsantos0101:
I've updated the "cycle" function. The change consists of a more straightforward application (it has been tested on both Ellie and localhost).

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -192,24 +192,14 @@ resulting list will be empty.
     --> [ 1, 2 ]
 
 -}
-cycle : Int -> List a -> List a
-cycle len list =
+cycle : Int -> List a -> List a 
+cycle c list =
     let
-        cycleLength =
-            List.length list
+        aux = List.length list
     in
-    if cycleLength == 0 || cycleLength == len then
-        list
-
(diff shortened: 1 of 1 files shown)


pzp1997:
Interesting implementation! The current implementation of `cycle` was written to be stack safe and performant, see #80. At the time, I tried several implementations and that was the best one I found, however, I do not believe I tried a "doubling" implementation like the one you are proposing. As per the [contribution guidelines](https://github.com/elm-community/list-extra#contributing), we need benchmarks that demonstrate performance gains over the current implementation to proceed.

pzp1997:
Going off intuition, I think you should avoid recomputing the length of the list for each recursive call since that results in many extra traversals of the list. Instead, you can compute the length of the list once, pass that as an argument to a recursive helper function, and add it to itself (double it) for each recursive call. 

Another change that I expect would improve perf is to stop the recursion _right before_ the length of the list is larger than `c` to avoid extra allocations.

flsantos0101:
Alright, I'll make some minor changes as suggested. Moreover, my apologies about the lack of benchmarking performance diagnosis, I'll provide it as well.

PR #130, opened 2019-11-11
Added oxfordJoin function

3digitdev:
elm auto-format cleaned up some stuff too

For reference, I'm using https://atom.io/packages/elm-format to format things.


3digitdev:
I just realized this is already a thing in `string-extra` package.  Nevermind...

PR #133, opened 2020-01-02
Fix #132

### LINKED ISSUE #132: isInfixOf returns incorrect result when a partial match is found before a complete match
In version 8.2.2
Small example, shown in repl:

```
> List.Extra.isInfixOf [1,2] [1,3,1,2]
False : Bool
```

It should return `True` .

The problem is that the implementation switches to `isPrefixOf` as soon as a single item matches.

pzp1997:
This PR addresses the logical error in `isInfixOf` discovered in #132. I added a unit test to prevent future regressions.

In the process of fixing the bug, I found a second unrelated issue in `isPrefixOf` that could cause a stack overflow for sufficiently large lists. (Since `isSuffixOf` and `isInfixOf` use `isPrefixOf` as a helper, they are also affected.) I added unit tests to make sure these functions do not cause a stack overflow in the future.

Lastly, I added some fuzz tests for `isPrefixOf`, `isSuffixOf`, and `isInfixOf`.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1688,7 +1688,7 @@ isInfixOfHelp infixHead infixTail list =
 
         x :: xs ->
             if x == infixHead then
-                isPrefixOf infixTail xs
+                isPrefixOf infixTail xs || isInfixOfHelp infixHead infixTail xs
 
             else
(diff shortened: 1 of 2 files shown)


Chadtech:
Awesome! Thank you.

PR #134, opened 2020-01-27
Added reverseRange, a faster way of creating a reversed range

JoshuaHall:
Added reverseRange, a faster way of creating a reversed range.
```elm
reverseRange 20 1 == List.range 1 20 |> List.reverse
```

Benchmarks (using [elm-explorations/benchmark](https://github.com/elm-explorations/benchmark)) on my machine show it is roughly 50% faster at most list sizes.

Benchmark comparing the naive implementation vs reverseRange

I also added a fuzz test for reverseRange. The range ceiling is there to prevent a heap error I was getting when using the `int` fuzzer.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1,9 +1,8 @@
-
 module List.Extra exposing
     ( last, init, getAt, uncons, unconsLast, maximumBy, maximumWith, minimumBy, minimumWith, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, setIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
     , intercalate, transpose, subsequences, permutations, interweave, cartesianProduct, uniquePairs
     , foldl1, foldr1, indexedFoldl, indexedFoldr
-    , scanl, scanl1, scanr, scanr1, mapAccuml, mapAccumr, unfoldr, iterate, initialize, cycle
+    , scanl, scanl1, scanr, scanr1, mapAccuml, mapAccumr, unfoldr, iterate, initialize, cycle, reverseRange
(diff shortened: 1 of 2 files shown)


Chadtech:
Hey @JoshuaHall ,

That is pretty cool!

What lead to you making this function? Did you run into a use case where you needed high performance reverse range? What was it, if you dont mind me asking?

JoshuaHall:
Hi, when I was forking [elm-chronographify](https://github.com/JoshuaHall/elm-chronographify/) (for educational purposes, I was originally learning about how to make a timer/stopwatch in Elm) I decided that it would make the code cleaner to implement this function. Otherwise I had to do a `List.reverse`, then a `List.indexedMap`, then finally a `List.reverse` again. With `reverseRange`, I could then implement my own `reversedIndexedMap`, except within it I operated over the `reverseRange` of the `List.length list` to `1`, not `(List.length list) - 1` to `0` (this removed the need to increment the index in the view, which again made the code slightly cleaner in my opinion).

The performance hasn't really come into play for `elm-chronogrpahify`, but I think for other apps that need something like this it could be very useful. However, the biggest benefit in my opinion is still the slightly shorter and clearer code.

prikhi (commented):
It ends up cleaner if you switch the arguments of the helper so that the `list` comes first, then you don't need to define arguments for the top-level function. Might as well use `high` & `low` instead of `hi` & `lo` as well:

```elm
reverseRange : Int -> Int -> List Int
reverseRange =
    let
        helper : List Int -> Int -> Int -> List Int
        helper list high low =
            if high >= low then
                helper (low :: list) high (low + 1)

            else
                list
    in
    helper []
```

JoshuaHall:
@prikhi I agree. I initially just copied `elm/core`'s `List.range` implementation and changed a few bits, but that seems nicer. I just committed it.

Chadtech:
I am still thinking about this. Right now I am leaning towards including this new function in `List.Extra`.

Usually I like better use case examples. Just having a shorter and simpler function in itself isn't all that compelling of a reason to include this. It doesnt encapsulate very much complexity that developers would struggle to implement on their own.

But it is optimized, and it's nice that `List.Extra` includes a lot of optimized versions of basic functions.

Chadtech:
Alright, I like it! Lets merge it. Thanks @JoshuaHall

PR #136, opened 2020-02-20
Add setBy

vjrasane:
A utility function for cases that Ive ran into few times now and have had to write it every time so I figured I might as well create a pull request :)

Most often when I use this I need to replace a value in a list by ID or name, so I'd call this by saying:

`setBy .id value list`

which is much nicer than:

`setIf (\item -> item.id == value.id) value list`

(or the shorter `setIf (.id >> (==) value.id) value list` )

Let me know if there are any problems with this or if I've missed some utility that achieves the same.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1,6 +1,6 @@
 
 module List.Extra exposing
-    ( last, init, getAt, uncons, unconsLast, maximumBy, maximumWith, minimumBy, minimumWith, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, setIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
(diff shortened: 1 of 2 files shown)


Chadtech:
Hey @vjrasane,

Do you have another use case for this function aside from setting something by id? Ids go along well with other data types like `Array` and `Dict`. So if the use case could be better served by those data types and the apis, then I would rather not include functions that make it easier to set specific values inside a `List` when it really seems like users should be using `Array` or `Dict`.

vjrasane:
The use cases all do revolve around *some* identifying value, of course. Quite often its an ID, but I've used it to set a user by their name, a bank account by its IBAN and a financial instrument by its ISIN. 

I think the key thing to recognize is that in some cases the order of items is important, so a dictionary would lose that information although it would make lookup by some comparable value easier. Same with Array. since the lookup is done by index. Another consideration is that you might need to set values based on several different values, so using a dictionary wouldn't help.

In any case I do see your point, although personally I have found this function to be useful on several occasions.

PR #143, opened 2021-03-27
Add reverseSortBy

Janiczek:
Alternative name might be `sortByDescending`?

Notably this implementation doesn't call `List.reverse`.


Chadtech:
Hey @Janiczek !

I made this benchmark https://ellie-app.com/cKKDPFjHsC9a1 and it looks like `List.sortBy identity >> List.reverse` is faster. But, since that is also the naive and obvious way of reversing a sort then maybe its not useful to have a helper for this.

Janiczek:
@Chadtech I haven't benchmarked this, good thinking!
If reverse-using variant is faster, let's not merge this then 👍💯

PR #144, opened 2021-04-08
Implement `mapMaybe` function

ntreu14:
I've implemented this a couple times in smaller personal projects of mine. I've seen this function implemented in the core libraries of [Haskell](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Maybe.html#v:mapMaybe) and [F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html#choose) (called `choose` in F#).

You can think of this function as doing a `filter` and `map` in one. Although this is a very contrived example, you an easily imagine one more complex:

```elm
[1, 2, 3, 4, 5]
  |> List.filter (\x -> x < 3)
  |> List.map (\x -> x * x)

-- Becomes:

perhapsSquareLessThanThree x =
  if x < 3
  then x * x |> Just
  else Nothing

mapMaybe perhapsSquareLessThanThree [1, 2, 3, 4, 5]
```

There are also other situation where you have a function that might return a `Maybe` and you can directly apply that function to a list of elements to only get back a list where it is `Just a`.

Let me know what you think. I'd love to hear thoughts!


Chadtech:
I love this function, but it is already in `elm/core` as `List.filterMap`.

ntreu14:
@Chadtech Wow, thanks for the heads up! I've never seen it called that before so didn't recognize it. I shoulda searched for the function signature.

Thanks for letting me know it already exists!

PR #147, opened 2021-06-03
Make gatherWith tail-call optimized

jfmengels:
This makes `gatherWith` tail-call optimized. I haven't benchmarked it, but since there is no other significant change in this function, I can only imagine that the performance improved.

Note: I discovered this issue with https://package.elm-lang.org/packages/jfmengels/elm-review-performance/latest/

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -2017,6 +2017,6 @@ gatherWith testFn list =
                         ( gathering, remaining ) =
                             List.partition (testFn toGather) population
                     in
-                    helper remaining <| ( toGather, gathering ) :: gathered
+                    helper remaining (( toGather, gathering ) :: gathered)
     in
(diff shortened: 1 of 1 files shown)


Chadtech:
Awesome! I did a benchmark just to check. In small lists, the performance increase appears to be around 5%: https://ellie-app.com/dnMGv4hW5kza1

jfmengels:
I need to get into the habit of benchmarking, thanks setting it up!

For posterity's sake, here is my benchmark results
- Brave (Version 1.25.68 Chromium: 91.0.4472.77 (Official Build) (64-bit))
- Ubuntu 20.04

Screenshot from 2021-06-05 23-14-03

So not as good as the 5% you're seeing, but definitely an improvement, and stack safety :+1:

PR #148, opened 2021-06-05
Make isInfixOf tail-call optimized

jfmengels:
- Added test to check that isInfixOf is not stack-safe. Without the change, the tests fails because of a stack overflow.
- Made isInfixOf tail-call optimized in all branches

Becnhmark: https://ellie-app.com/dnQMpprLrDPa1

Screenshot from 2021-06-05 23-32-04

Note: There are still quite a few functions that can be optimized in this package, but I likely won't propose PRs for these, as they become a bit more complex.

You can run the following to find them:
```bash
npx elm-review --template jfmengels/elm-review-performance/example --rules NoUnoptimizedRecursion
```

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1742,7 +1742,11 @@ isInfixOfHelp infixHead infixTail list =
 
         x :: xs ->
             if x == infixHead then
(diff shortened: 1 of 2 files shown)


Chadtech:
Awesome! Thanks again

PR #151, opened 2021-08-19
remove Set usage in unique functions, drop comparable restriction

cjoach:
The comparable type variable is restrictive in the form that it doesn't work with custom types. For the family of `unique` functions, only the `==` operator is needed so comparable is not strictly needed. The restriction seemed to come from using `Set` to containing unique elements instead of a list.

I removed the restriction, please tell me if I'm overlooking something.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -414,16 +414,16 @@ dropWhile predicate list =
     --> [ 0, 1 ]
 
 -}
-unique : List comparable -> List comparable
+unique : List a -> List a
 unique list =
-    uniqueHelp identity Set.empty list []
+    uniqueHelp identity [] list []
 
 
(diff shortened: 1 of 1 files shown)


Chadtech:
Awesome discovery. It seems plausible that this was made this way for a good reason, tho I dont know what that reason is.

I made this benchmark (https://ellie-app.com/f5Kzb3J2cx9a1) and it seems that the `List` implementation is faster for about 100 element long lists, and at around 150 it becomes slower. Quite a lot depends on the composition of the list, and the function used to determine uniqueness.

Seems like a good trade off. I am not sure. I'll think about it and come back to this later.

Chadtech:
Alright. Lets do it. 40% Better performance on the range of 0-100 sized `List`s and being able to get uniqueness on any kind of element is worth the 30% drop in performance on >200 sized `List`s.

Chadtech:
Thanks @cjoach

PR #152, opened 2021-08-22
better performing `isPermutationOf`

lue-bird:
`isPermutationOf` is implemented as
```elm
isPermutationOf permut xs =
    member permut (permutations xs)
```
It first calculates all **n! permutations**, then looks if the given list matches any. This makes `isPermutationOf` a very heavy operation:
grafik
[ellie][ellie]

We can do the same job in O(n^2) time:
```elm
isPermutationOf permut xs =
    (length permut == length xs)
        && (permut |> all (\a -> xs |> member a))
```
You could also remove elements you already checked from the second list but in effect, both implementations perform equally as good.
grafik
[ellie][ellie]

The PR also adds tests and examples for the new `isPermutationOf` and fixes a spelling mistake (applyed) in the `allDifferentBy` doc.

[ellie]: https://ellie-app.com/f5pWDmnpmhva1

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1807,10 +1807,7 @@ In other words: Do the 2 `List`s contain the same elements but in a different or
 isPermutationOf : List a -> List a -> Bool
 isPermutationOf permut xs =
     (length permut == length xs)
-        && (permut
-                |> List.all
-                    (\a -> xs |> member a)
-           )
+        && (permut |> all (\a -> xs |> member a))
 
 
 {-| Take two lists and returns a list of corresponding pairs
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1786,6 +1786,23 @@ isSubsequenceOf subseq list =
 
 
 {-| Take two lists and return `True`, if the first list is a permutation of the second list.
+In other words: Do the 2 `List`s contain the same elements but in a different order?
+
+    [ 3, 1, 2 ]
+        |> isPermutationOf
+            [ 1, 2, 3 ]
+    --> True
+
+    [ 3, 1, 0 ]
+        |> isPermutationOf
+            [ 1, 2, 3 ]
+    --> False
+
+    [ 3, 1, 2, 2 ]
+        |> isPermutationOf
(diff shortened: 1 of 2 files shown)


Chadtech:
Awesome. That is a huge increase! Thank you

gampleman:
This is indeed much faster, unfortunately it is also incorrect.

```
> isPermutationOf [1,1,1] [1,2,3]
True : Bool
```

I would suggest reverting this.

---

Also, there is an easy oracle based test that can be used for any new implementation:

```elm
fuzz2 (list int) (list int) "works the same as sorting" <|
 \(a b) ->
    isPermutationOf a b
       |> Expect.equal (List.sort a == List.sort b)
```

lue-bird:
Sorry for the trouble! Thanks @gampleman for noticing this!
A version that works (hopefully) and has the same performance:
```elm
isPermutationOf : List a -> List a -> Bool
isPermutationOf permut xs =
    let
        removeOneMember : a -> List a -> { foundAny : Bool, without : List a }
        removeOneMember culprit =
            removeOneMemberHelp culprit []

        removeOneMemberHelp culprit before list =
            case list of
                [] ->
                    { foundAny = False, without = [] }

                head :: after ->
                    if head == culprit then
                        { foundAny = True, without = before ++ after }

                    else
                        removeOneMemberHelp culprit (head :: before) after
    in
    case ( permut, xs ) of
        ( [], [] ) ->
            True

        ( [], _ :: _ ) ->
            False

        ( _ :: _, [] ) ->
            False

        ( _ :: _, x :: after ) ->
            let
                { foundAny, without } =
                    removeOneMember x permut
            in
            if foundAny then
                isPermutationOf without after

            else
                False
```

Chadtech:
Thanks all for catching this so quickly. I plan on circling back to this tonight and releasing a new version.

Chadtech:
Okay. I released `8.5.1` with your new implementation @lue-bird . It passed the tests, passed a new test that `[1,1,1]` is not a permutation of `[1,2,3]`, and I added @gampleman 's fuzz test.

Thanks again!

PR #154, opened 2021-09-15
isPermutationOf: Add an edge-case the fuzz test missed

Janiczek:
The edge case actually helped disprove the validity of the implementation of `isPermutationOf` that I had in my head:

```elm
isPermutationOf xs ys =
  (List.length xs == List.length ys)
    && List.all (\x -> List.member x ys) xs
    && List.all (\y -> List.member y xs) ys
```


Chadtech:
Great! Thanks @Janiczek

PR #155, opened 2021-09-18
isPermutationOf performance increase

Chadtech:
We just merged in some big performance gains for `isPermutationOf` from @lue-bird , and since then @jfmengels gave some tips on how to further increase the performance. Namely: (0) dont curry arguments, (1) define all helper functions at the top level, and (2) Re-arrange case and if statements so that the most common cases are branched to earlier.

I followed his advice and I saw performance gains between 270% and 340% depending on the inputs

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1824,40 +1824,17 @@ In other words: Do the 2 `List`s contain the same elements but in a different or
 
     [ 3, 1, 2, 2 ]
         |> isPermutationOf
-            [ 1, 2, 3 ]
+            [ 1, 2, 3 ]r
     --> False
 
 -}
 isPermutationOf : List a -> List a -> Bool
 isPermutationOf permut xs =
-    let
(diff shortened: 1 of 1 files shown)


Chadtech:
Pinging @gampleman @Janiczek too, just in case you are interested as well.

src/List/Extra.elm
```diff
@@ -1824,40 +1824,17 @@ In other words: Do the 2 `List`s contain the same elements but in a different or
 
     [ 3, 1, 2, 2 ]
         |> isPermutationOf
-            [ 1, 2, 3 ]
+            [ 1, 2, 3 ]r
```
Janiczek: Probably a typo :)

Chadtech: Thanks!

gampleman:
LGTM. 

Also I wonder if a note about the complexity of the algorithm isn't in order? By my squint, this is still quadratic, right? I think having a little warning about things that can grow their runtime steeply may well be in order.

PR #157, opened 2021-12-14
Make groupsOf and greedyGroupsOf fn families tail-recursive

### LINKED ISSUE #75: Group methods will crash if used with a large list

Janiczek:
Closes #75

I've hit this during Advent of Code 😭 😄

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1864,33 +1864,31 @@ skipped and not appear in any groups.
 -}
 groupsOfWithStep : Int -> Int -> List a -> List (List a)
 groupsOfWithStep size step list =
-    let
-        go : List a -> List (List a) -> List (List a)
-        go xs acc =
-            if List.isEmpty xs then
-                List.reverse acc
+    if size <= 0 || step <= 0 then
+        []
 
-            else
-                let
-                    thisGroup =
-                        List.take size xs
-
-                    rest =
-                        List.drop step xs
-
-                    okayArgs =
-                        size > 0 && step > 0
-
-                    okayLength =
-                        size == List.length thisGroup
-                in
-                if okayArgs && okayLength then
-                    go rest (thisGroup :: acc)
+    else
+        let
+            go : List a -> List (List a) -> List (List a)
+            go xs acc =
+                if List.isEmpty xs then
+                    List.reverse acc
 
                 else
-                    go rest acc
-    in
-    go list []
+                    let
+                        thisGroup =
+                            List.take size xs
+
+                        rest =
+                            List.drop step xs
+                    in
+                    if size == List.length thisGroup then
+                        go rest (thisGroup :: acc)
+
+                    else
+                        List.reverse acc
+        in
+        go list []
 
 
 {-| `groupsOfVarying ns` takes `n` elements from a list for each `n` in `ns`, splitting the list into variably sized segments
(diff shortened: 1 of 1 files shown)


src/List/Extra.elm
```diff
@@ -1863,25 +1863,34 @@ skipped and not appear in any groups.
...
-    else
-        []
+                    okayLength =
+                        size == List.length thisGroup
+                in
+                if okayArgs && okayLength then
+                    go rest (thisGroup :: acc)
+
+                else
+                    go rest acc
```
jfmengels: Can't we skip one iteration by calling `List.reverse acc` in this case?

src/List/Extra.elm
```diff
@@ -1863,25 +1863,34 @@ skipped and not appear in any groups.
...
+                    rest =
+                        List.drop step xs
 
-        okayLength =
-            size == List.length thisGroup
-    in
-    if okayArgs && okayLength then
-        thisGroup :: groupsOfWithStep size step xs_
+                    okayArgs =
+                        size > 0 && step > 0
```
jfmengels: This can probably be computed once in `groupsOfWithStep` but outside of `go`

src/List/Extra.elm
```diff
@@ -1863,25 +1863,32 @@ skipped and not appear in any groups.
...
+                        thisGroup =
+                            List.take size xs
 
-        okayLength =
-            size == List.length thisGroup
-    in
-    if okayArgs && okayLength then
-        thisGroup :: groupsOfWithStep size step xs_
+                        rest =
+                            List.drop step xs
```
jfmengels: You could move this definition to inside the if (then) condition. That way, we won't compute the rest if `size /= List.length thisGroup`.

Janiczek: Done in fb47d4a, thanks!

Chadtech:
Thanks for looking into this. Sorry for the delay, but I will check out this PR ASAP (later today or tomorrow).

Chadtech:
I just benchmarked this here: https://ellie-app.com/gm9X8yfPLXMa1 (with a one small adjustment: I pulled `go` out into a top level function). There seems to be a performance drop primarily on small lists and small group sizes. 

(Ignore the fact that its labelled "isPermutationOf". Thats because I copy and pasted the "isPermutationOf" benchmark to get started).
Screen Shot 2022-01-08 at 02 56 25
Screen Shot 2022-01-08 at 02 56 37
Screen Shot 2022-01-08 at 02 56 52

Is it worth it? Probably. But what do you all think?

Janiczek:
> Is it worth it?

I'd consider it worth it, as correctness issues IMHO should have more priority than performance issues

jfmengels:
I agree with @Janiczek, correctness is more important than performance (at least as a generally available function like this). In a way, it's good that it's slower mostly for the small inputs because performance increases for those don't matter too much in the big picture anyway (I mean, you can still call this function 2 million times per second apparently!).

Chadtech:
Sounds good to me. Thanks @Janiczek !

PR #158, opened 2021-12-18
updateAt: Avoid computing head when unnecessary

jfmengels:
This makes it so that we don't unnecessarily compute `head` in the case where the tail is empty.


Chadtech:
Makes sense. Thanks @jfmengels !

PR #159, opened 2021-12-18
Make remove tail-call optimized

jfmengels:
This makes `remove` tail-call optimized, meaning it will not blow up the stack when working on lists that are too large.

|This also improves performance:](https://github.com/jfmengels/elm-benchmarks/blob/master/src/ImprovingPerformance/ListExtra/Remove.elm)

Screenshot from 2021-12-18 10-50-05

I also special-cased the case where the element is not found so that we don't unnecessarily recreate the entire list from the accumulator.

I also added tests for `remove` and applied elm-format (seems like tests weren't properly formatted).

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -854,16 +854,21 @@ updateIfIndex predicate update list =
 -}
 remove : a -> List a -> List a
 remove x xs =
+    removeHelp xs x xs []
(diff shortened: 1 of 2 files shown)


Chadtech:
Awesome work! Thanks @jfmengels

PR #160, opened 2022-01-14
List slicing and enumerate Functions

threecifanggen:


Chadtech:
For `enumerate`, lets not include that in `List.Extra` just because its simple. Thats my opinion at least.

For `islice`, what does "islice" mean? And then, more importantly, what is it useful for? Is there a project where this was needed? What was the use case?

threecifanggen:
> For `enumerate`, lets not include that in `List.Extra` just because its simple. Thats my opinion at least.
> 
> For `islice`, what does "islice" mean? And then, more importantly, what is it useful for? Is there a project where this was needed? What was the use case?

The idea `islice` is from Python [`itertoolz.islice`](https://docs.python.org/3/library/itertools.html#itertools.islice), I find it is useful and basic for data manipulation(because I am a data scientist, I want to visualize some data through an elm project).

Chadtech:
I guess we don't have any kind of `List` slicing function at all in the core list module or in `List.Extra`. There is a `slice` function in `Array`. Would an `Array` work in your data visualization use case? Or would there be a problem with doing that? Do you do any other `Array` like things like get or set values by index?

threecifanggen:
> I guess we don't have any kind of `List` slicing function at all in the core list module or in `List.Extra`. There is a `slice` function in `Array`. Would an `Array` work in your data visualization use case? Or would there be a problem with doing that? Do you do any other `Array` like things like get or set values by index?

Yep. What I want is what you say.

PR #162, opened 2022-02-23
Add stoppable folds

Janiczek:
Sometimes these functions come in handy (and would allow beginners to stay in the fold world instead of being forced to write their own general recursion functions).

Prior art: 
* [Clojure](https://clojuredocs.org/clojure.core/reduced)
* [Gleam](https://hexdocs.pm/gleam_stdlib/gleam/list.html#fold_until)

> Slightly off-topic: I've had bad experience making the code doc examples run with `elm-verify-examples` manually. (And I found out `elm-verify-examples`-generated tests don't even run in Travis here 🙈 )

Screenshot 2022-02-23 at 21 32 19

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1,7 +1,7 @@
 module List.Extra exposing
     ( last, init, getAt, uncons, unconsLast, maximumBy, maximumWith, minimumBy, minimumWith, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, setIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
     , intercalate, transpose, subsequences, permutations, interweave, cartesianProduct, uniquePairs
-    , Step(..), foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl, stoppableFoldr
+    , Step(..), foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl
     , scanl, scanl1, scanr, scanr1, mapAccuml, mapAccumr, unfoldr, iterate, initialize, cycle, reverseRange
     , splitAt, splitWhen, takeWhileRight, dropWhileRight, span, break, stripPrefix, group, groupWhile, inits, tails, select, selectSplit, gatherEquals, gatherEqualsBy, gatherWith
     , isPrefixOf, isSuffixOf, isInfixOf, isSubsequenceOf, isPermutationOf
@@ -27,7 +27,7 @@ module List.Extra exposing
 
 # Folds
 
-@docs Step, foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl, stoppableFoldr
+@docs Step, foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl
 
 
 # Building lists
@@ -1291,25 +1291,6 @@ stoppableFoldl func acc list =
 
                 Stop finalAcc ->
                     finalAcc
-
-
-{-| A `foldr` that can stop early instead of traversing the whole list.
-
-    stoppableFoldr
-        (\n acc ->
-            if acc >= 50 then
-                Stop acc
-            else
-                Continue (n + acc)
-        )
-        0
-        (List.range 1 10000)
-    --> 10000
-
--}
-stoppableFoldr : (a -> b -> Step b) -> b -> List a -> b
-stoppableFoldr func acc list =
-    stoppableFoldl func acc (List.reverse list)
 
 
 {-| Reduce a list from the left, building up all of the intermediate results into a list.
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -1,7 +1,7 @@
 module List.Extra exposing
     ( last, init, getAt, uncons, unconsLast, maximumBy, maximumWith, minimumBy, minimumWith, andMap, andThen, reverseMap, takeWhile, dropWhile, unique, uniqueBy, allDifferent, allDifferentBy, setIf, setAt, remove, updateIf, updateAt, updateIfIndex, removeAt, removeIfIndex, filterNot, swapAt, stableSortWith
     , intercalate, transpose, subsequences, permutations, interweave, cartesianProduct, uniquePairs
-    , foldl1, foldr1, indexedFoldl, indexedFoldr
+    , Step(..), foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl, stoppableFoldr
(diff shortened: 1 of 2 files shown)


src/List/Extra.elm
```diff
@@ -27,7 +27,7 @@ module List.Extra exposing
 
 # Folds
 
-@docs foldl1, foldr1, indexedFoldl, indexedFoldr
+@docs Step, foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl, stoppableFoldr
```
jfmengels: `Step` whould be better placed right before the stoppable functions I'd say.

tests/Tests.elm
```diff
@@ -975,4 +975,49 @@ all =
...
+                        )
+                        0
+                        (List.range 1 1000)
+                        |> Expect.equal 55
+            ]
+        , describe "stoppableFoldr"
+            [ fuzz (list int) "behaves like stoppableFoldl with reversed list" <|
+                \xs ->
+                    stoppableFoldr
+                        (\n acc ->
```
jfmengels: To make it clearer and/or less error-prone, you could put the function inside a let declaration, so that it's very clear that both operations do the same thing.

src/List/Extra.elm
```diff
@@ -1244,6 +1244,74 @@ indexedFoldr func acc list =
...
+        x :: xs ->
+            case func x acc of
+                Continue newAcc ->
+                    stoppableFoldl func newAcc xs
+
+                Stop finalAcc ->
+                    finalAcc
+
+
+{-| A `foldr` that can stop early instead of traversing the whole list.
```
jfmengels: But it does traverse the whole list because of the `reverse` :smiling_face_with_tear:

jfmengels: I don't think there are a lot of performance comments in this package, but it may be worth mentioning that if the whole list ends up being visited, that this version will be slower than `List.foldr`. What do you think?

Chadtech:
Hey @Janiczek ! Sorry you had a bad experience with the travis set up and the tests. It seems like that is something we need to clean up.

Usually stuff we merge into `List.Extra` is either satisfying some need or use case, or a performance improvement.

As far as performance, lets make a benchmark. Here is one I made for a recent PR:  https://ellie-app.com/gm9X8yfPLXMa1 (I can get around to testing this out sometime within ~1 week if no one else wants to make it).

As far as the use case. I am not sure about this. So I can see the code examples in the documentation..
```elm
    stoppableFoldr
        (\n acc ->
            if acc >= 50 then
                Stop acc
            else
                Continue (n + acc)
        )
        0
        (List.range 1 10000)
 ```
 ..but the non-`List.Extra` alternative of what is above would be..
 ```elm
    List.foldr
        (\n acc ->
            if acc >= 50 then
                acc
            else
                n + acc
        )
        0
        (List.range 1 10000)
 ```
..which is simpler in my view (less stuff, less types, less characters, fewer dependencies..).

Maybe theres something else? I think I remember someone doing something similar when the fold function was very complex and the `acc` value was a big record. It helped make a complex code path more explicit.

What are your thoughts?

**Comment by jfmengels** [comment]
I made a benchmark here: https://github.com/jfmengels/elm-benchmarks/blob/master/src/WhatIsFaster/StoppableFold.elm
and these are the results.

### Chrome



### Firefox



### Safari

File: src/List/Extra.elm (hunk 3 above)
Comment (Janiczek) [reply src/List/Extra.elm]: I agree with the comments needing to change if we keep the implementation as it is.

I was wondering whether to try doing what List.foldr is doing (with its threshold of 500*3 items or whatever that magic is). We could also not provide this stoppableFoldr function at all?

Comment (jfmengels) [reply src/List/Extra.elm]: I think both are possible solutions. Not providing foldr is definitely the easiest one.

**Comment by Janiczek** [comment]
Thanks to feedback from both of you I've removed the `stoppableFoldr` function. I previously added it without much thought and it's true the List.reverse inside totally removes the benefit.

**Comment by Chadtech** [comment]
Wow thats quite a performance boost. Thank you @jfmengels for the benchmarks. Since the performance gain is so much, lets include it.

Another question, what do you think about a `Maybe` instead of `Step`? I know `Step` is more descriptive and that is really nice, but doesnt the addition of a type add some complexity? I wouldnt think its much of an issue if this `List.Extra` was a small module, but since its very big, its hard to see just from the api what `Step` would be related to.

**Comment by jfmengels** [comment]
WiTh Step, you can define the value to return when using Stop, or to use for the next iteration using Continue. With a Maybe, you'd be missing one of those in the Nothing case. So I don't think it's applicable.

**Comment by Chadtech** [comment]
Oh Im sorry of course. `Maybe a` has a different shape than `Step a`.

What about `(a, Bool)`? Where `True` represents continue?

**Comment by jfmengels** [comment]
I think a boolean won't be as easy to understand, as there will be an implicit meaning for a boolean value. It can be especially confusing if your entire function is returning a boolean, in which case the function passed to `stoppableFoldl` will need to return a `( Bool, Bool )`, in which case I can easily imagine switching the 2.

Having `Stop`/`Continue` also has the benefit of being to pipe a value in those variants, instead of wrapping the entire expression in a tuple.

```elm
stoppableFoldl (\x acc ->
  case x of
    A a ->
      a
        |> f
        |> g
        |> Continue
    
    B b ->
      b
        |> f
        |> g
        |> Stop
)
```

The boolean does have the benefit of reducing the API and making it super obvious how to go from a boolean value to a stop condition (although you may use `not` to use the right one, which might be confusing as well).

I think that `Step` is the better approach here.

src/List/Extra.elm
```diff
@@ -27,7 +27,7 @@ module List.Extra exposing
 
 # Folds
 
-@docs foldl1, foldr1, indexedFoldl, indexedFoldr
+@docs Step, foldl1, foldr1, indexedFoldl, indexedFoldr, stoppableFoldl
```
jfmengels: I think it's better to have the type defined as close as possible to the function that uses it. Otherwise they'll have to CTRL-F the other element when looking at the type/function.

```suggestion
@docs foldl1, foldr1, indexedFoldl, indexedFoldr, Step, stoppableFoldl
```

jfmengels:
I still think we should mention the performance characteristics, as this is likely important in a bunch of cases.

Chadtech:
I agree re documenting performance characteristics.

What if we put this in its own module? Like `List.Extra.Step`? That way the `Step` type wont be buried among a large api and it can be the main focus of its module?

jfmengels:
I don't think it warrants its own module. It is still almost an implementation detail of the new function. Putting the type right next to the function where it will be used should be just fine IMO. I do it a bunch in my packages and I haven't heard about people finding it confusing.

Chadtech:
Alright. I really like this one. Let's do it. Thanks for your patience all.

PR #166, opened 2022-07-15
Add a stricter test for stoppableFoldl behaviour

Janiczek:
@adamnfish has found that the tests for `List.Extra.stoppableFoldl` aren't really testing the main promise of the function: that it will stop running the function after a `Stop` is returned.

Together we're adding a test which will throw a RangeError (stack overflow) if `stoppableFoldl` goes through the whole list. We've verified it by temporarily changing the implementation of `stoppableFoldl` to just be a `List.foldl`.

### DIFF
--- a/tests/Tests.elm
+++ b/tests/Tests.elm
@@ -993,5 +993,30 @@ all =
                         0
                         (List.range 1 1000)
                         |> Expect.equal 55
+            , test "after Stop, the function is not called anymore" <|
+                let
+                    throwRangeErrorException : () -> a
+                    throwRangeErrorException () =
+                        preventTailCallOptimization ()
+
+                    preventTailCallOptimization : () -> a
+                    preventTailCallOptimization () =
+                        throwRangeErrorException ()
+                in
+                \() ->
+                    stoppableFoldl
+                        (\n acc ->
+                            if n < 50 then
+                                Continue n
+
+                            else if n == 50 then
+                                Stop n
+
+                            else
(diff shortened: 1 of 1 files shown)


Chadtech:
Nice catch! Thank you.

tests/Tests.elm
```diff
@@ -993,5 +993,30 @@ all =
...
+                    stoppableFoldl
+                        (\n acc ->
+                            if n < 50 then
+                                Continue n
+
+                            else if n == 50 then
+                                Stop n
+
+                            else
+                                explode ()
```
Janiczek, line 1016: @Chadtech I've just realized this probably is missing a function name change 😰  `explode` can't be found anywhere, if my memory's correct it was supposed to be `throwRangeErrorException ()` instead of `explode ()`. This will likely not compile.

Sorry, I guess a commit got lost / didn't get pushed / ... I'll create a fix PR

Janiczek: Actually the [last version of the code](https://github.com/elm-community/list-extra/blob/master/tests/Tests.elm#L1131) has `throwRangeErrorException ()` already. Is this some kind of GitHub glitch? Weird. (EDIT: ah, it was fixed in #163) I guess we don't need to do anything then

Chadtech: No worries! I noticed the code wasn't compiling after I merged it in, but it was an easy fix.

Thanks for your original contribution, and thanks for noticing now!

PR #169, opened 2022-12-02
Speed up minimumBy and maximumBy

jfmengels:
This PR speeds up these 2 functions.

The change is that we don't recreate tuples unnecessary, which does have a cost.

I made [some benchmarks](https://github.com/jfmengels/elm-benchmarks/blob/master/src/ImprovingPerformance/ListExtra/MinimumBy.elm) (only for `minimumBy`, but it's the same for `maximumBy`).


Chadtech:
Beautiful! Thank you @jfmengels

PR #172, opened 2023-01-11
Optimize List.Extra.init (changing the type signature)

Janiczek:
This uses `length` and `take` on the list instead of two `reverse`s.

Benchmark: https://gist.github.com/Janiczek/67a106dbfdd18d029f99522868bd5126

EDIT: I forgot to pass `items` to the second algorithm! 🤦🤦🤦🤦🤦 

Here are the new, correct benchmarks:

Safari:

Screenshot 2023-01-11 at 18 45 51

Chrome:

Screenshot 2023-01-11 at 18 46 06

I guess it doesn't make sense to merge this now.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -100,23 +100,15 @@ last items =
 {-| Return all elements of the list except the last one.
 
     init [ 1, 2, 3 ]
-    --> Just [ 1, 2 ]
+    --> [ 1, 2 ]
 
     init []
-    --> Nothing
+    --> []
 
 -}
-init : List a -> Maybe (List a)
+init : List a -> List a
 init items =
-    case items of
-        [] ->
-            Nothing
-
-        nonEmptyList ->
-            nonEmptyList
-                |> List.reverse
-                |> List.tail
-                |> Maybe.map List.reverse
+    List.take (List.length items - 1) items
 
 
 {-| Returns `Just` the element at the given index in the list,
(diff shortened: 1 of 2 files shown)


src/List/Extra.elm
```diff
@@ -100,23 +100,15 @@ last items =
...
-    case items of
-        [] ->
-            Nothing
-
-        nonEmptyList ->
-            nonEmptyList
-                |> List.reverse
-                |> List.tail
-                |> Maybe.map List.reverse
+    List.take (List.length items - 1) items
```
jfmengels, line 111: Why not keep the signature as it was? The empty check should be pretty quick.

Janiczek: Yeah, if the original signature is preferable I can put it back. Honestly I like the `List.take` / `List.drop` kind of API, where it returns a List, more.

jfmengels: I think it's better. A breaking change to this package (which so many other depends on) is not going to be fun to see :sweat_smile:

Chadtech:
I searched github for instances of `List.Extra.init` and it seems like of the ~15 results, ~75% of them were piped into `Maybe.withDefault []`. That said it did seem like a minority of cases use the `Nothing` case. Interestingly, I noticed multiple cases where `List.Extra.init` and `List.Extra.last` were used together, as if people want to separate out the last element like `uncons` does for the first element. Maybe `List.Extra` needs `unconsLast`?

PR #175, opened 2023-02-09
Add function: isPermutationOfWith

annaghi:
Added a `With` variant for `isPermutationOf`.


annaghi:
Closing this PR, because I realized that I can convert the lists to `Dict` and use `Dict.diff` twice on them.

PR #176, opened 2023-06-29
Add consecutivePairs and pairwise

Janiczek:
```elm
consecutivePairs : List a -> List ( a, a )
pairwise : (a -> a -> b) -> List a -> List b

consecutivePairs [ 1, 2, 3, 4 ]
--> [ ( 1, 2 ), ( 2, 3 ), ( 3, 4 ) ]

pairwise (-) [100, 10, 1]
--> [ 90, 9 ]
```


gampleman:
Not really sure that `consecutivePairs` is really worth the space is takes given that it's trivially `pairwise Tuple.pair`. I think there was some good wisdom in the design of the core libraries when `map2` was preferred over Haskell's `zip`...

Chadtech:
Do we have some use cases associated with these? @Janiczek did you run into anything recently that lead you to using functions like this

I like `pairwise` and I think I have been in the situation where I wanted that function.

But for `consecutivePairs`, my hunch is that no one wants `(a, a)` as the final step in whatever they are doing. Getting consecutive elements paired up is always an intermediate step in something bigger, so it isn't clear why it needs to be tupled up to begin with. Does that make sense?

Janiczek:
Makes sense! We can probably go without `consecutivePairs`. I was mostly copying the existing API of `uniquePairs : List a -> List (a,a)`.

My hunch is that I run into the need for `pairwise` a few times a year. This PR was prompted by one of these but I can't remember what it was for :joy:

PR #177, opened 2023-07-11
Add `reverseFilter`, following the style of `reverseMap`

nibrivia:
Similarly to `reverseMap`, adds an tail-call optimized, but reversing `filter` operation: `reverseFilter`.


Chadtech:
This looks great. Thank you! I ran some benchmarks here and I can see it is more performant: https://ellie-app.com/nn24pGfpq82a1

PR #182, opened 2024-08-13
add splitOn

ggPeti:
I'm rude. My code is good. Accept the PR for the code or reject it for the person's rudeness. The choice is yours.

### DIFF
--- a/src/List/Extra.elm
+++ b/src/List/Extra.elm
@@ -10,6 +10,7 @@ module List.Extra exposing
     , lift2, lift3, lift4
     , groupsOf, groupsOfWithStep, groupsOfVarying, greedyGroupsOf, greedyGroupsOfWithStep
     , joinOn
+    , splitOn
     )
 
 {-| Convenience functions for working with List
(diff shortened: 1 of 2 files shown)


Janiczek:
For reviewers: the intuitive sense as to why this doesn't need any `List.reverse` nor `++` is that instead of tail-optimization it explodes the calculation into many nested function calls.

@ggPeti I think it could be useful to mention in the doc comment for the function that this keeps the delimiters

miniBill:
Doesn't this cause a stack overflow for very long lists?

ggPeti:
@Janiczek thanks, isn't "starting with each element" clear in communicating that? that was my intention

@miniBill yes it does. It's a purposefully stack-utilizing recursive function.

miniBill:
I think one of the requisites we have for `-extra` functions is that they are stack safe.

By the way, the PR should probably be against `elmcraft/core-extra`.

ggPeti:
I refuse to submit PRs to `elmcraft/core-extra`. And requiring stack safety indiscriminately is just silly and no such guarantee is documented in this library.

miniBill:
Acknowledged

ggPeti:
Hi @Chadtech could you please approve the checks?

